PR #19, opened 2026-04-13
feat(sns): add SubscriptionBuilder

### LINKED ISSUE #27: Support SNS Subscriptions
SNS package currently supports Topics. Add support for Subscription

laazyj:
## Summary

Adds `createSubscriptionBuilder()` to `@composurecdk/sns`, generalising `AWS::SNS::Subscription` across every SNS protocol (email, email-json, HTTPS/HTTP, SMS, SQS, Lambda, Firehose, application). Motivated by a user porting a budget → SNS → Lambda auto-shutdown CloudFormation template onto composureCDK — previously the only option was dropping to raw CDK for the subscription.

Follows the established `TopicBuilder` pattern and `docs/architecture.md`:

- `Lifecycle`-compatible, proxied fluent API via `Builder`.
- `Resolvable<ITopic>` and `Resolvable<IQueue>` so `topic` and `deadLetterQueue` can be wired via `ref()` inside `compose()` — no post-build imperative wiring.
- Build-time validation for required `topic`, `protocol`, and `endpoint` with descriptive errors.
- Build result surfaces the `Subscription` construct.

## Recommended alarms live on `TopicBuilder`

`NumberOfNotificationsRedrivenToDlq` and `NumberOfNotificationsFailedToRedriveToDlq` are topic-level metrics (dimension `TopicName` only), so creating them per-subscription would produce duplicate alarms when subscriptions share a topic. Both are now added to `TopicBuilder`'s recommended alarms with `TreatMissingData: notBreaching`, so topics without any DLQ-enabled subscriptions stay silent.

| Alarm                                       | Metric                                                   | Default threshold |
| ------------------------------------------- | -------------------------------------------------------- | ----------------- |
| `numberOfNotificationsRedrivenToDlq`        | `NumberOfNotificationsRedrivenToDlq` (Sum, 1 min)        | `> 0`             |
| `numberOfNotificationsFailedToRedriveToDlq` | `NumberOfNotificationsFailedToRedriveToDlq` (Sum, 1 min) | `> 0`             |

## Subscription defaults

Intentionally none. The PR previously shipped `rawMessageDelivery: false` as a "pinned" default, but that's also CDK's default — it was noise, not safety. Protocol-specific defaults (e.g. enforcing HTTPS) are worth a follow-up design pass per the review feedback.

Deliberately not defaulted:

- **Dead-letter queue** — the primary reliability knob per the AWS Well-Architected Reliability Pillar and [SNS DLQ guidance](https://docs.aws.amazon.com/sns/latest/dg/sns-dead-letter-queues.html), but requires a caller-owned `Queue`. A future `@composurecdk/sqs` `QueueBuilder` would unlock auto-creation; tracked separately.
- **HTTPS-over-HTTP enforcement** — allowed but called out in the README; matches `TopicBuilder`'s precedent of leaving transport choices to the caller.

## Test plan

- [x] `npx nx run sns:typecheck` — passes.
- [x] `npx nx run sns:test` — 40 tests pass.
- [x] `npx nx run-many -t test` — all packages green (example snapshots regenerated for the 2 new topic alarms).
- [x] `npm run format:check` — clean.
- [x] `npm run lint` — clean.
- [x] End-to-end: test case composes `createTopicBuilder()` + `createSubscriptionBuilder()` via `ref(...)` and asserts synthesized CloudFormation contains one `AWS::SNS::Topic` and one `AWS::SNS::Subscription` wired to it — covering the originating CFN YAML shape.

Closes #27

https://claude.ai/code/session_01U6hVYWJWmC46EzpX8F6mK7


packages/sns/src/index.ts
```diff
@@ -6,3 +6,12 @@ export {
 export { TOPIC_DEFAULTS } from "./defaults.js";
 export { type TopicAlarmConfig } from "./alarm-config.js";
 export { TOPIC_ALARM_DEFAULTS } from "./alarm-defaults.js";
+export {
+  createSubscriptionBuilder,
+  type ISubscriptionBuilder,
+  type SubscriptionBuilderProps,
```
laazyj: Why expose `SubscriptionBuilderProps`? This is used for generating the fluent builder interface. I don't see why a consumer of the package would need to use this type.

packages/sns/src/subscription-alarm-config.ts
```diff
@@ -0,0 +1,53 @@
...
+ * only in that case. When no DLQ is configured, `createSubscriptionAlarms`
+ * returns an empty record regardless of this configuration.
+ *
+ * Set individual alarms to `false` to disable them, or provide an
+ * {@link AlarmConfig} to tune thresholds.
+ *
+ * @see https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Best_Practice_Recommended_Alarms_AWS_Services.html#SNS
+ * @see https://docs.aws.amazon.com/sns/latest/dg/sns-dead-letter-queues.html
+ */
+export interface SubscriptionAlarmConfig {
```
laazyj: Both of these alarms appear to be related to the `Topic` rather than the `Subscription`. Should they be recommended alarms on the `TopicBuilder` rather than the `SubscriptionBuilder`?

packages/sns/src/subscription-alarms.ts
```diff
@@ -0,0 +1,117 @@
...
+import type { AlarmDefinition } from "@composurecdk/cloudwatch";
+import { createAlarms, resolveAlarmConfig } from "@composurecdk/cloudwatch";
+import type { SubscriptionAlarmConfig } from "./subscription-alarm-config.js";
+import { SUBSCRIPTION_ALARM_DEFAULTS } from "./subscription-alarm-defaults.js";
+
+const METRIC_PERIOD = Duration.minutes(1);
+const METRIC_PERIOD_LABEL = `${String(METRIC_PERIOD.toMinutes())} minute`;
+const SNS_NAMESPACE = "AWS/SNS";
+
+function topicMetric(topic: ITopic, metricName: string): Metric {
```
laazyj: As above, this is starting to smell. The alarms are defined _per-topic_ not _per-subscription_. They are in the wrong builder.

packages/sns/src/subscription-builder.ts
```diff
@@ -0,0 +1,181 @@
...
+  /**
+   * Dead-letter queue for messages that cannot be delivered to the
+   * subscribed endpoint. Accepts a concrete {@link IQueue} or a
+   * {@link Ref} to another component's output.
+   *
+   * Attaching a DLQ is the primary reliability control for SNS
+   * subscriptions and also enables the recommended redrive alarms.
+   *
+   * @see https://docs.aws.amazon.com/sns/latest/dg/sns-dead-letter-queues.html
+   * @default - no dead letter queue
```
laazyj, line 37: What does AWS Well Architected Framework say about adding DLQs to SNS Subscriptions? Should we consider spinning up a queue by default (with the user able to disable/override this behaviour)? That would allow/require the following: 
1. an `sqs` package with `QueueBuilder` to build a well-architected, secure, SQS queue for the DLQ.
2. AWS recommended alarms for a DLQ (need to distinguish DLQ from another Queue type).

laazyj (commented):
Good overall — the implementation is faithful to the `TopicBuilder` pattern and the test coverage is thorough. A few things worth addressing before merge:

**🔴 Double `enabled` check (inconsistency with `TopicBuilder`)**

`createSubscriptionAlarms` checks `config?.enabled` (line 322–323) and then passes `config` into `resolveSubscriptionAlarmDefinitions`, which re-checks `config?.enabled === false` (line 251). `createTopicAlarms` only checks `enabled` once (in the outer function). Recommend removing the `enabled` check from `resolveSubscriptionAlarmDefinitions` so the outer function owns that gate, matching the existing pattern.

**🟡 Duplicate alarms when multiple subscriptions share a topic**

Metrics are constructed with `dimensionsMap: { TopicName: topic.topicName }`, which is topic-level. If two subscriptions share a topic, they'll each create a separate alarm construct monitoring the *exact same metric* — they'll always fire together and are functionally identical. The README mentions metrics are topic-level but doesn't warn users about this. Worth adding a note along the lines of: *"If multiple subscriptions share a topic, each will create its own alarm construct monitoring the same topic-level metric. Consider creating the alarm once at the topic level instead."*

**🟡 No build-time validation for `protocol` and `endpoint`**

`topic` gets a descriptive error if missing, but `protocol` and `endpoint` don't — they'll produce a generic CDK synth error. Worth adding explicit validation for consistency with the `topic` check.

**🟡 No `addAlarm()` method**

`TopicBuilder` exposes `addAlarm()` for custom alarms; `SubscriptionBuilder` doesn't. Either add it for parity, or explicitly document in the README that custom alarms aren't supported on subscriptions.

**🟢 Minor: `METRIC_PERIOD_LABEL` pluralisation**

`${String(METRIC_PERIOD.toMinutes())} minute` would produce `"5 minute"` if the period were ever changed. Pre-existing issue in `topic-alarms.ts` too, so not a blocker here.

packages/sns/src/subscription-defaults.ts
```diff
@@ -0,0 +1,27 @@
+import type { SubscriptionProps } from "aws-cdk-lib/aws-sns";
+
+/**
+ * Secure, AWS-recommended defaults applied to every SNS subscription built
+ * with {@link createSubscriptionBuilder}. Each property can be individually
+ * overridden via the builder's fluent API.
+ *
+ * SNS subscription defaults are deliberately minimal because the meaningful
+ * defaults are protocol-specific (e.g. raw message delivery is invalid for
+ * email/SMS but often desirable for SQS/Lambda). Reliability is best
```
laazyj: This might need further exploration. What would it look like to provide _protocol-specific_ defaults so that we can provide more secure/reliable out-of-the-box infrastructure with minimal configuration?

laazyj (commented):
Requires further work,

laazyj:
Thanks for the thorough review. Addressed in ddd6712 — the structural feedback about the alarms being topic-level turned out to resolve most of the other points too:

**🔴 Double `enabled` check** — resolved by removing the subscription-level alarm module entirely. The remaining path (`createTopicAlarms`) already checks `enabled` only once.

**🟡 Duplicate alarms across subscriptions sharing a topic** — resolved by moving the alarms to `TopicBuilder`. A topic now gets exactly one redrive alarm + one failed-redrive alarm regardless of how many subscriptions it has.

**🟡 Missing `protocol`/`endpoint` validation** — added with matching error messages (`SubscriptionBuilder "Sub": protocol is required…` / `endpoint is required…`) and tests.

**🟡 No `addAlarm()` method on SubscriptionBuilder** — left off intentionally, since there are no subscription-level SNS CloudWatch metrics. Custom subscription-related alarms should go on the `TopicBuilder` via its existing `.addAlarm()`. README updated to reflect this.

**🟢 `METRIC_PERIOD_LABEL` pluralisation** — agreed, out of scope; happy to fix separately across both SNS and any other builders using the same pattern if it's worth doing.

packages/sns/src/subscription-alarm-config.ts
laazyj: Agreed — both metrics have only a `TopicName` dimension, so they're unambiguously topic-level. Moved to `TopicAlarmConfig` / `TOPIC_ALARM_DEFAULTS` / `topic-alarms.ts`, with `TreatMissingData: notBreaching` so topics whose subscriptions don't have DLQs stay quiet. `subscription-alarm-config.ts`, `subscription-alarm-defaults.ts`, and `subscription-alarms.ts` are gone.

packages/sns/src/subscription-alarms.ts
laazyj: Yep — file deleted, alarm definitions moved into `topic-alarms.ts` as additional entries alongside `numberOfNotificationsFailed` and `numberOfNotificationsFilteredOutInvalidAttributes`. The alarm IDs are stable (keyed on the topic's scope) and there's only ever one of each per topic now.

packages/sns/src/subscription-builder.ts
laazyj: Strong agree that a default DLQ is the Well-Architected answer, and splitting it out makes sense:

1. `@composurecdk/sqs` with a `QueueBuilder` for a well-architected, secure SQS queue.
2. A `DlqQueueBuilder` variant (or a role/tag on `QueueBuilder`) with the DLQ-specific recommended alarms (`ApproximateAgeOfOldestMessage`, `ApproximateNumberOfMessagesVisible`, etc. — these are queue-level, unlike the SNS redrive metrics which are topic-level).
3. `SubscriptionBuilder` auto-materialises one unless the caller passes their own queue / opts out.

That's a meaningful surface-area addition and feels like a follow-up PR rather than in-scope for this one. Happy to open a tracking issue for it if that's useful — for now this PR keeps DLQs caller-owned and documents the Well-Architected recommendation in the README.

packages/sns/src/subscription-defaults.ts
laazyj: Worth a design pass on its own. A sketch of what it could look like:

- A `SubscriptionProtocol` discriminant selects a protocol-specific `Partial<SubscriptionProps>` default block.
- Examples: SQS → `rawMessageDelivery: true` by default (more useful for downstream consumers, and valid for this protocol); HTTP → hard error / warn unless explicitly opted in, and maybe an auto-DLQ; Lambda → `rawMessageDelivery: true` plus permission wiring; Firehose → raw delivery + an IAM role hookup.
- The global `SUBSCRIPTION_DEFAULTS` from this PR (just `rawMessageDelivery: false`) was more noise than safety — it pinned CDK's own default. I've removed it in the push; protocol-specific defaults seem like the better frame to bring back under.

Agree this is a follow-up rather than a blocker for landing the basic builder. Happy to open a tracking issue.

packages/sns/src/index.ts
laazyj: Good catch — removed. `SubscriptionBuilderProps` is now only used internally (it was also missing the `-internal` rule from the other builders). Index exports only `createSubscriptionBuilder`, `ISubscriptionBuilder`, and `SubscriptionBuilderResult`.

packages/sns/src/subscription-builder.ts
laazyj: Tracked as #34.

packages/sns/src/subscription-defaults.ts
laazyj: Tracked as #35.

packages/examples/test/__snapshots__/dual-function-app.test.ts.snap
```diff
@@ -68,6 +94,32 @@ exports[`dual-function-app > matches the expected synthesised template 1`] = `
       },
       "Type": "AWS::CloudWatch::Alarm",
     },
+    "DualFunctionAppalertsNumberOfNotificationsRedrivenToDlqAlarm434BC23F": {
```
laazyj, line 97: I'm curious why the example snapshots changed to add DLQ alarms, since the example app does not specify a DLQ?

laazyj: Because I added both DLQ redrive alarms to TOPIC_ALARM_DEFAULTS, so every topic built with createTopicBuilder() gets them regardless of whether any subscription has a DLQ. The
  justification was TreatMissingData: notBreaching — the alarm stays OK when the metric has no data — but on a topic with no DLQ-attached subscriptions the alarm is functionally inert,
  just noise (and ~$0.10/mo each in CloudWatch).

PR #20, opened 2026-04-14
feat(iam): add RoleBuilder, ManagedPolicyBuilder, and StatementBuilder

laazyj:
Introduce @composurecdk/iam with three fluent builders that centralise the
IAM primitives the Lambda, Budgets, and SNS-topic-policy packages all need:

- createRoleBuilder() — Lifecycle builder for iam.Role with a default
  one-hour maxSessionDuration (Well-Architected SEC pillar), Resolvable
  permissionsBoundary for cross-component wiring, and
  addInlinePolicyStatements() that accepts raw PolicyStatements or
  StatementBuilders (resolved at build time).
- createManagedPolicyBuilder() — Lifecycle builder for customer-managed
  ManagedPolicy with addStatements().
- createStatementBuilder() — fluent wrapper around PolicyStatement, not a
  Lifecycle, with a least-privilege guardrail that throws
  WildcardResourceError for Effect=Allow + resources ["*"] unless the
  caller opts in via .allowWildcardResources(true).
- createServiceRoleBuilder(service) — thin sugar over
  createRoleBuilder().assumedBy(new ServicePrincipal(...)).

Defaults and alarms follow the conventions used by the existing SNS, S3,
and Lambda packages. Full test coverage via vitest + CDK Template
assertions.

https://claude.ai/code/session_01RiAH14trizgmHhaRGd5BRz

### DIFF
--- a/packages/budgets/src/budget-builder.ts
+++ b/packages/budgets/src/budget-builder.ts
@@ -1,0 +1,268 @@
+import { CfnBudget, type CfnBudgetProps } from "aws-cdk-lib/aws-budgets";
+import { type Alarm } from "aws-cdk-lib/aws-cloudwatch";
+import type { TopicPolicy } from "aws-cdk-lib/aws-sns";
+import type { IConstruct } from "constructs";
+import { Builder, type IBuilder, type Lifecycle } from "@composurecdk/core";
+import type { BudgetAlarmConfig } from "./alarm-config.js";
+import { createBudgetAlarms } from "./budget-alarms.js";
+import { BUDGET_DEFAULTS } from "./defaults.js";
+import {
+  type BudgetSubscriber,
+  type NotificationEntry,
+  type NotificationType,
+  resolveSubscribers,
+  toCfnNotificationWithSubscribers,
+} from "./notifications.js";
+import { createBudgetsTopicPolicies } from "./topic-policy.js";
+
+/**
+ * Spend limit for a cost or usage budget.
+ */
+export interface BudgetLimit {
+  /** Numeric limit. */
+  amount: number;
+  /**
+   * Currency or usage unit. Defaults to
+   * {@link BUDGET_DEFAULTS.limitUnit} when omitted.
+   */
+  unit?: string;
+}
+
+/**
+ * Configuration properties for the budget builder.
+ */
+export interface BudgetBuilderProps {
+  /** Name used for `BudgetName` in the `BudgetData` property. */
+  budgetName?: string;
+  /**
+   * One of `COST`, `USAGE`, `RI_UTILIZATION`, `RI_COVERAGE`,
+   * `SAVINGS_PLANS_UTILIZATION`, `SAVINGS_PLANS_COVERAGE`.
+   *
+   * @default "COST"
+   */
+  budgetType?: string;
+  /** @default "MONTHLY" */
+  timeUnit?: string;
+  /** Spend limit (required for COST and USAGE budgets). */
+  limit?: BudgetLimit;
+  /**
+   * CloudFormation `CostFilters` map. Keys are filter dimensions
+   * (`Service`, `Region`, `LinkedAccount`, `TagKeyValue`, …) and values
+   * are arrays of filter values.
+   */
+  costFilters?: Record<string, string[]>;
+  /** CloudFormation `CostTypes` passthrough. */
+  costTypes?: CfnBudget.CostTypesProperty;
+  /**
+   * Account that owns the budget. Defaults to the account the stack is
+   * deployed into; passing an explicit value is useful only for payer
+   * accounts managing member-account budgets.
+   */
+  billingAccountId?: string;
+  /**
+   * Configuration for the opt-in billing alarm.
+   * @see BudgetAlarmConfig
+   */
+  recommendedAlarms?: BudgetAlarmConfig | false;
+}
+
+/**
+ * The build output of an {@link IBudgetBuilder}.
+ */
+export interface BudgetBuilderResult {
+  /** The `AWS::Budgets::Budget` construct. */
+  budget: CfnBudget;
+  /**
+   * `AWS::SNS::TopicPolicy` constructs created automatically for any SNS
+   * topic referenced as a notification subscriber, keyed by the topic's
+   * node id. Grants `budgets.amazonaws.com` permission to publish.
+   *
+   * `{}` when no SNS subscribers were configured.
+   */
+  topicPolicies: Record<string, TopicPolicy>;
+  /**
+   * CloudWatch alarms created for the budget (currently only the
+   * optional `estimatedCharges` billing alarm). Empty unless the caller
+   * opts in via `recommendedAlarms`.
+   */
+  alarms: Record<string, Alarm>;
+}
+
+/**
+ * A fluent builder for configuring and creating an AWS Budget.
+ *
+ * Wraps the {@link CfnBudget} L1 construct (the CDK does not ship an L2
+ * for Budgets) with well-architected defaults, helpers for the
+ * percentage-threshold notification shape, and automatic
+ * `AWS::SNS::TopicPolicy` wiring for SNS subscribers.
+ *
+ * @example
+ * ```ts
+ * createBudgetBuilder()
+ *   .budgetName("AgentBudget")
+ *   .limit({ amount: 50, unit: "GBP" })
+ *   .notifyOnActual(100, ref("alerts", r => r.topic))
+ *   .withRecommendedThresholds()
+ *   .build(stack, "AgentBudget");
+ * ```
+ */
+export type IBudgetBuilder = IBuilder<BudgetBuilderProps, BudgetBuilder>;
+
+class BudgetBuilder implements Lifecycle<BudgetBuilderResult> {
+  props: Partial<BudgetBuilderProps> = {};
+  private readonly _notifications: NotificationEntry[] = [];
+  private _recommendedThresholdsApplied = false;
+  private _recommendedThresholdSubscribers: BudgetSubscriber[] = [];
+
+  /**
+   * Add a notification that fires when ACTUAL spend crosses the given
+   * percentage of the budget limit.
+   *
+   * @param thresholdPercent - Percentage of the budget limit (e.g. `80`).
+   * @param subscribers - One or more email addresses / SNS topics (or
+   *   {@link Resolvable} refs to topics).
+   */
+  notifyOnActual(thresholdPercent: number, ...subscribers: BudgetSubscriber[]): this {
+    return this._addPercentageNotification("ACTUAL", thresholdPercent, subscribers);
+  }
+
+  /**
+   * Add a notification that fires when FORECASTED spend crosses the
+   * given percentage of the budget limit.
+   */
+  notifyOnForecasted(thresholdPercent: number, ...subscribers: BudgetSubscriber[]): this {
+    return this._addPercentageNotification("FORECASTED", thresholdPercent, subscribers);
+  }
+
+  /**
+   * Raw notification passthrough for callers that need the full
+   * CloudFormation surface (e.g. absolute-value thresholds).
+   */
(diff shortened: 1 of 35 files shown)


packages/iam/src/index.ts
```diff
@@ -0,0 +1,20 @@
+export {
+  createRoleBuilder,
+  type IRoleBuilder,
+  type RoleBuilderProps,
```
laazyj, line 4: Do not export internal props that are used solely for creating the fluent builder interface.

packages/iam/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  type RoleBuilderResult,
+} from "./role-builder.js";
+export { ROLE_DEFAULTS } from "./role-defaults.js";
+export {
+  createManagedPolicyBuilder,
+  type IManagedPolicyBuilder,
+  type ManagedPolicyBuilderProps,
```
laazyj, line 11: Do not export internal props that are used solely for creating the fluent builder interface.

packages/budgets/src/budget-builder.ts
```diff
@@ -0,0 +1,268 @@
...
+    const timeUnit = budgetProps.timeUnit ?? BUDGET_DEFAULTS.timeUnit;
+
+    const requiresLimit = budgetType === "COST" || budgetType === "USAGE";
+    if (requiresLimit && !budgetProps.limit) {
+      throw new Error(
+        `BudgetBuilder "${id}": limit({ amount, unit }) must be set for ${budgetType} budgets.`,
+      );
+    }
+
+    this._applyRecommendedThresholds();
```
laazyj, line 176: **Bug: `_applyRecommendedThresholds()` mutates `_notifications` on every `build()` call.**

If `build()` is called more than once on the same builder instance, the recommended thresholds are pushed into `_notifications` again, duplicating them. Fix: apply the thresholds once at configuration time (inside `withRecommendedThresholds`) rather than lazily here, or work on a local snapshot:

```ts
// Option A: move the push into withRecommendedThresholds()
withRecommendedThresholds(...subscribers: BudgetSubscriber[]): this {
  const { actualPercent, forecastedPercent } = BUDGET_DEFAULTS.recommendedThresholds;
  this._notifications.push(
    { notificationType: "ACTUAL", thresholdPercent: actualPercent, subscribers },
    { notificationType: "FORECASTED", thresholdPercent: forecastedPercent, subscribers },
  );
  return this;
}
```

laazyj (commented):
Overall this is a well-structured PR — consistent with existing patterns, well-documented, and solid test coverage. A couple of bugs need fixing before merge; the rest are naming/consistency issues.

packages/budgets/src/budget-builder.ts
```diff
@@ -0,0 +1,268 @@
...
+            },
+          }
+        : {}),
+      ...(budgetProps.costFilters ? { costFilters: budgetProps.costFilters } : {}),
+      ...(budgetProps.costTypes ? { costTypes: budgetProps.costTypes } : {}),
+    };
+
+    const cfnProps: CfnBudgetProps = {
+      budget: budgetData,
+      ...(budgetProps.billingAccountId ? { billingAccountId: budgetProps.billingAccountId } : {}),
```
laazyj, line 198: **Bug: `billingAccountId` is not a valid field on `CfnBudgetProps`.**

`CfnBudgetProps` has no `billingAccountId` property — that field is `accountId` on `CfnBudget.BudgetDataProperty`. This spread is silently ignored at runtime. It should be added to `budgetData` instead:

```ts
const budgetData: CfnBudget.BudgetDataProperty = {
  ...
  ...(budgetProps.billingAccountId ? { accountId: budgetProps.billingAccountId } : {}),
};
```

packages/budgets/src/topic-policy.ts
```diff
@@ -0,0 +1,52 @@
...
+export function createBudgetsTopicPolicies(
+  scope: IConstruct,
+  id: string,
+  topics: ITopic[],
+): Record<string, TopicPolicy> {
+  const policies: Record<string, TopicPolicy> = {};
+  const seen = new Set<string>();
+
+  for (const topic of topics) {
+    const key = topic.node.id;
```
laazyj, line 30: **Fragile deduplication key — can silently drop topic policies.**

`topic.node.id` is only unique within a parent scope. Two topics in different scopes can share the same `node.id` (e.g. both named `"AlertsTopic"`), causing the second topic's policy to be silently skipped — leaving it without the required `budgets.amazonaws.com` publish permission.

Use `topic.node.path` instead, which is fully-qualified and always unique within a CDK app:

```ts
const key = topic.node.path;
```

Note: the `topicPolicies` result map in `BudgetBuilderResult` is also documented as "keyed by the topic's node id" — that doc comment would need updating too.

packages/iam/src/role-builder.ts
```diff
@@ -0,0 +1,174 @@
...
+    };
+
+    const role = new Role(scope, id, mergedProps);
+
+    const inlinePolicies: Record<string, Policy> = {};
+    for (const entry of this._inlinePolicies) {
+      const resolvedStatements = entry.statements.map((s) =>
+        s instanceof StatementBuilder ? s.build() : s,
+      );
+      const policy = new Policy(scope, `${id}Policy${entry.name}`, {
```
laazyj, line 147: **`addInlinePolicyStatements` is a misnomer — these are attached policies, not inline policies.**

`new Policy(...).attachToRole(role)` creates an `AWS::IAM::Policy` resource and attaches it to the role. In CloudFormation terms this is a *managed inline policy* — distinct from the `Policies` array embedded directly in `AWS::IAM::Role` (which is what CDK's `inlinePolicies` prop on `RoleProps` produces).

This matters because:
- Attached policies count against the 10-policy-per-role limit separately.
- The CloudFormation resource type differs (`AWS::IAM::Policy` vs embedded in `AWS::IAM::Role`).

Suggestion: rename to `addPolicyStatements` / `attachedPolicies`, or switch the implementation to pass statements via the `inlinePolicies` prop on `RoleProps` to produce truly inline policies.

packages/sns/src/subscription-builder.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    return { subscription };
+  }
+}
+
+/**
+ * Creates a new {@link SubscriptionBuilder} for attaching a subscription to
+ * an externally-built topic.
+ */
+export function createSubscriptionBuilder(): SubscriptionBuilder {
```
laazyj, line 85: **Inconsistent with every other builder — not wrapped with `Builder()`.**

Every other `create*Builder()` function in the codebase returns `Builder(SomeClass)`, which wraps the class in the fluent proxy and exports an opaque `I*Builder` type. This one returns `new SubscriptionBuilder()` directly and exports the concrete class type.

Since `SubscriptionBuilder` uses private fields rather than a `props` object, the `Builder` proxy wouldn't add prop setters — but the inconsistency in the exported type is still worth addressing. Consider either:
- Documenting why this builder intentionally doesn't use the `Builder` wrapper, or
- Moving `_topic` / `_subscription` into a `props` object and wrapping with `Builder()`, exporting an `ISubscriptionBuilder` type alias.

packages/sns/src/topic-builder.ts
```diff
@@ -103,7 +147,13 @@ class TopicBuilder implements Lifecycle<TopicBuilderResult> {
 
     const alarms = createTopicAlarms(scope, id, topic, alarmConfig, this.customAlarms);
 
-    return { topic, alarms };
+    const subscriptions: Record<string, Subscription> = {};
+    for (const entry of this._subscriptions) {
+      const resolvedSub = resolve(entry.subscription, context);
+      subscriptions[entry.key] = topic.addSubscription(resolvedSub);
```
laazyj, line 153: **`ITopic.addSubscription()` returns `void`, not `Subscription`.**

The CDK type definition for `ITopic.addSubscription()` returns `void`. Assigning it to `subscriptions[entry.key]` (typed as `Record<string, Subscription>`) is a type error that would surface at compile time without `skipLibCheck`.

The `Subscription` construct needs to be created explicitly, similar to how `SubscriptionBuilder.build()` does it:

```ts
const subscriptionConfig = resolvedSub.bind(topic);
subscriptions[entry.key] = new Subscription(scope, `${id}${entry.key}`, {
  topic,
  ...subscriptionConfig,
});
```

packages/budgets/src/notifications.ts
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * Callers that need the full CloudFormation surface (custom comparison
+ * operators, absolute-value thresholds) should use this shape; for the
+ * common percentage case, prefer
+ * {@link IBudgetBuilder.notifyOnActual} /
+ * {@link IBudgetBuilder.notifyOnForecasted}.
+ */
+export interface NotificationEntry {
+  notificationType: NotificationType;
+  thresholdPercent: number;
```
laazyj, line 31: **`thresholdPercent` is a misleading name when `thresholdType` is `ABSOLUTE_VALUE`.**

When `thresholdType: "ABSOLUTE_VALUE"` is used (e.g. via `addNotification`), this value is a dollar amount, not a percentage. The CloudFormation property is simply called `threshold`. Suggest renaming to `threshold` and updating the percentage helpers (`notifyOnActual`, `notifyOnForecasted`) to document that they pass a percentage value.

packages/iam/src/managed-policy-defaults.ts
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Defaults applied to every customer-managed policy built with
+ * {@link createManagedPolicyBuilder}.
+ *
+ * Customer-managed policies are intentionally light on defaults — the
+ * permissive surface of a managed policy is determined entirely by the
+ * statements the caller adds, and guardrails against overly permissive
+ * statements live in {@link StatementBuilder}.
+ */
+export const MANAGED_POLICY_DEFAULTS: Partial<ManagedPolicyProps> = {};
```
laazyj, line 12: **Empty defaults object adds public API surface for no current value.**

The SNS and Lambda packages only export a defaults file when it contains actual values. An empty `{}` here creates a file, an export, and a public API entry point with nothing in it. Suggest removing this until there's a meaningful default to add (e.g. `path: "/"`).

packages/iam/src/index.ts
laazyj: Addressed in https://github.com/laazyj/composureCDK/pull/36

packages/iam/src/index.ts
laazyj: Addressed in https://github.com/laazyj/composureCDK/pull/36

packages/iam/src/role-builder.ts
laazyj: Addressed in https://github.com/laazyj/composureCDK/pull/36

packages/iam/src/managed-policy-defaults.ts
laazyj: Addressed in https://github.com/laazyj/composureCDK/pull/36

packages/sns/src/subscription-builder.ts
laazyj: https://github.com/laazyj/composureCDK/pull/39

packages/budgets/src/budget-builder.ts
laazyj: See https://github.com/laazyj/composureCDK/pull/40

packages/budgets/src/topic-policy.ts
laazyj: See https://github.com/laazyj/composureCDK/pull/40

packages/sns/src/topic-builder.ts
laazyj: See https://github.com/laazyj/composureCDK/pull/39

packages/budgets/src/notifications.ts
laazyj: See https://github.com/laazyj/composureCDK/pull/40

Issue #38, opened 2026-04-18
SubscriptionBuilder skips IAM/resource-policy wire-up for Lambda/SQS endpoints


laazyj:
## Summary

The standalone \`createSubscriptionBuilder()\` in \`@composurecdk/sns\` constructs an \`AWS::SNS::Subscription\` directly from a \`protocol\` + \`endpoint\` pair. This skips the endpoint-specific IAM/resource-policy wire-up that CDK's \`ITopicSubscription.bind(topic)\` performs.

## Impact

- **Lambda endpoint:** no \`AWS::Lambda::Permission\` is added, so SNS cannot invoke the function — deliveries are rejected with \`AccessDenied\`.
- **SQS endpoint:** no \`AWS::SQS::QueuePolicy\` is added, so SNS cannot \`SendMessage\` — deliveries fail. (For KMS-encrypted queues, the Decrypt/GenerateDataKey statement is also missing.)
- **Email / HTTPS / URL endpoints:** unaffected — these don't require resource policies on the subscriber.

## Repro

\`\`\`ts
createSubscriptionBuilder()
  .topic(topic)
  .protocol(SubscriptionProtocol.LAMBDA)
  .endpoint(fn.functionArn)
  .build(stack, \"Sub\");
// Synthesised template has no AWS::Lambda::Permission granting sns.amazonaws.com
// → SNS → Lambda deliveries will fail at runtime.
\`\`\`

## Reference

CDK's [\`LambdaSubscription.bind()\`](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-sns-subscriptions/lib/lambda.ts) calls \`fn.addPermission(...)\`; [\`SqsSubscription.bind()\`](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/aws-sns-subscriptions/lib/sqs.ts) calls \`queue.addToResourcePolicy(...)\` and may add a KMS grant.

## Suggested fix

Adopt the \`ITopicSubscription\`/\`bind()\` pattern — the same approach #39 uses for the topic-builder \`addSubscription\` addition. Either:

1. **Replace** the \`protocol\`/\`endpoint\` API with a single \`.subscription(ITopicSubscription)\` setter (breaking). Cleanest.
[...]

PR #39, opened 2026-04-18
feat(sns): add TopicBuilder.addSubscription via ITopicSubscription.bind()

laazyj:
## Summary

Ports the one genuinely valuable addition from the SNS commit in #20 to the version of the SNS package that shipped via #19: co-located subscription declarations on `TopicBuilder`.

- `TopicBuilder.addSubscription(key, subscription)` accepts any `ITopicSubscription` (or `Resolvable<ITopicSubscription>`) and binds it via `ITopicSubscription.bind(topic)` — the same path CDK's own `topic.addSubscription(...)` uses, so endpoint-specific IAM/resource-policy wire-up happens automatically.
- `TopicBuilderResult.subscriptions: Record<string, Subscription>` exposes each created subscription, keyed by the name passed to `addSubscription` (`{}` when none).
- No change to `TOPIC_DEFAULTS`, recommended alarms, or the standalone `SubscriptionBuilder`.

## Differences from #20's implementation

- #20's reviewer flagged that `ITopic.addSubscription()` returns `void`, not `Subscription` — that implementation was relying on `skipLibCheck` to compile. This PR instead constructs the `Subscription` explicitly via `new Subscription(scope, id, { topic, ...sub.bind(topic) })`, which is the fix the reviewer suggested.
- The standalone `SubscriptionBuilder` from #20 is not ported here — main's `Builder()`-wrapped, protocol/endpoint-based `SubscriptionBuilder` (from #19) already addresses the other #20 feedback point (missing `Builder` wrapper).

## Known follow-up

Filed #38 for a separate IAM-correctness gap in the standalone `SubscriptionBuilder`: when targeting Lambda/SQS with a raw `protocol` + `endpoint`, the required `AWS::Lambda::Permission` / `AWS::SQS::QueuePolicy` are not emitted. This PR sidesteps that by giving the common case (co-located subscriptions) a correct `bind()`-based path; the fix for the standalone builder is a breaking-change decision worth handling on its own.

## Test plan

- [x] `npx nx test sns` (46/46 pass, 6 new)
- [x] `AWS::SNS::Subscription` created for `EmailSubscription`
- [x] `AWS::Lambda::Permission` (Action: `lambda:InvokeFunction`, Principal: `sns.amazonaws.com`) emitted for `LambdaSubscription`
- [x] `AWS::SQS::QueuePolicy` (Allow `sqs:SendMessage`, Principal: `sns.amazonaws.com`) emitted for `SqsSubscription`
- [x] `Resolvable<ITopicSubscription>` resolves from the compose context
- [x] `npm run lint`, `npm run format:check`

### DIFF
--- a/packages/sns/README.md
+++ b/packages/sns/README.md
@@ -126,6 +126,45 @@ for (const alarm of Object.values(result.alarms)) {
 }
 ```
 
+## Adding Subscriptions to a Topic
+
+For the common case where a topic and its subscriptions are declared together, use `addSubscription` on the topic builder. It accepts any CDK [`ITopicSubscription`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_sns.ITopicSubscription.html) (e.g. `EmailSubscription`, `LambdaSubscription`, `SqsSubscription`) and binds it via `ITopicSubscription.bind(topic)` — the same path CDK uses for `topic.addSubscription(...)`, so endpoint-specific wire-up (Lambda invoke permission, SQS queue policy, KMS decrypt policy) happens automatically.
+
+```ts
+import { createTopicBuilder } from "@composurecdk/sns";
+import { EmailSubscription, LambdaSubscription } from "aws-cdk-lib/aws-sns-subscriptions";
+
+const result = createTopicBuilder()
+  .topicName("alerts")
+  .addSubscription("ops", new EmailSubscription("ops@example.com"))
+  .addSubscription("handler", new LambdaSubscription(alertHandler))
+  .build(stack, "Alerts");
+
+result.subscriptions.ops; // AWS SNS Subscription construct
+```
+
+Each subscription is exposed on `result.subscriptions` under the key supplied to `addSubscription`.
+
+Cross-component subscriptions can be declared with `ref(...)` so the subscription's endpoint is resolved from another component's build output:
+
+```ts
+import { compose, ref } from "@composurecdk/core";
+import { createTopicBuilder } from "@composurecdk/sns";
+import { createFunctionBuilder, type FunctionBuilderResult } from "@composurecdk/lambda";
+import { LambdaSubscription } from "aws-cdk-lib/aws-sns-subscriptions";
+
+const system = compose(
+  {
+    handler: createFunctionBuilder()./* ... */,
+    alerts: createTopicBuilder().addSubscription(
+      "handler",
+      ref("handler", (r: FunctionBuilderResult) => new LambdaSubscription(r.function)),
+    ),
+  },
+  { handler: [], alerts: ["handler"] },
+);
+```
+
 ## Subscription Builder
 
 ```ts
--- a/packages/sns/README.md
+++ b/packages/sns/README.md
@@ -145,7 +145,7 @@ result.subscriptions.ops; // AWS SNS Subscription construct
 
 Each subscription is exposed on `result.subscriptions` under the key supplied to `addSubscription`.
 
-Cross-component subscriptions can be declared with `ref(...)` so the subscription's endpoint is resolved from another component's build output:
+Cross-component subscriptions can be declared with `ref(...)` inside a [`compose`](../core/README.md) system, so the subscription's endpoint is resolved from another component's build output at build time:
 
 ```ts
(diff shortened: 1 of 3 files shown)


laazyj:
Reviewer comments, already addressed.

## Code Review: PR #39 - feat(sns): add TopicBuilder.addSubscription via ITopicSubscription.bind()

### Summary
This PR adds co-located subscription declarations to TopicBuilder, allowing subscriptions to be defined alongside topics using CDK's ITopicSubscription interface. The implementation correctly uses bind() to ensure
proper IAM/resource-policy wire-up.

### Strengths

1. Correct IAM wire-up: The implementation uses ITopicSubscription.bind(topic) and explicitly constructs Subscription objects, which properly addresses the type-safety issue from the previous attempt (where
ITopic.addSubscription() returns void).

2. Comprehensive test coverage: The tests verify:
   - Basic subscription creation (EmailSubscription)
   - Lambda permission wire-up (AWS::Lambda::Permission)
   - SQS queue policy wire-up (AWS::SQS::QueuePolicy)
   - Resolvable support for cross-component references
   - Result structure (subscriptions keyed by name)

3. Good documentation: The README examples are clear and show both simple and cross-component use cases.

4. Consistent with project patterns: The addSubscription method follows the same fluent builder pattern as addAlarm.

### Issues & Suggestions

#### 1. Context parameter signature inconsistency

The build method signature differs from other builders in the codebase:

typescript
// topic-builder.ts (this PR)
build(scope: IConstruct, id: string, context: Record<string, object> = {}): TopicBuilderResult

// rest-api-builder.ts (existing)
build(scope: IConstruct, id: string, context?: Record<string, object>): RestApiBuilderResult

// function-builder.ts (existing)
build(scope: IConstruct, id: string): FunctionBuilderResult


The context parameter should be optional (context?: Record<string, object>) to match:
- The Lifecycle<T, Context> interface definition
- The rest-api-builder.ts implementation
- The principle that builders should work standalone without compose context

Fix:
typescript
build(scope: IConstruct, id: string, context?: Record<string, object>): TopicBuilderResult {
  // ...
  for (const entry of this._subscriptions) {
    const resolvedSub = resolve(entry.subscription, context ?? {});
    // ...
  }
}


#### 2. Subscription construct ID generation

The subscription ID is generated as ${id}${entry.key}:

typescript
subscriptions[entry.key] = new Subscription(scope, `${id}${entry.key}`, {


This creates IDs like TestTopicops (no separator). Consider:
- Adding a separator: ${id}Subscription${entry.key} → TestTopicSubscriptionops
- Or capitalizing the key: ${id}${capitalize(entry.key)} → TestTopicOps

The current approach works but could lead to less readable CloudFormation logical IDs.

Suggested fix:
typescript
subscriptions[entry.key] = new Subscription(scope, `${id}Subscription${entry.key}`, {


#### 3. Empty subscriptions object always present

The implementation always returns subscriptions: {} even when no subscriptions are added. While this is documented and tested, consider whether this is the right default:

Current:
typescript
const result = createTopicBuilder().build(stack, "Topic");
result.subscriptions; // {}


Alternative (optional property):
typescript
subscriptions?: Record<string, Subscription>;


The current approach is probably better for consistency (similar to how alarms is always present), but it's worth confirming this matches the project's conventions.

#### 4. Type safety for subscription keys

The addSubscription method accepts any string as a key, which could lead to collisions or typos. Consider whether the builder should:
- Validate for duplicate keys
- Throw an error if the same key is used twice

Example:
typescript
addSubscription(key: string, subscription: Resolvable<ITopicSubscription>): this {
  if (this._subscriptions.some(s => s.key === key)) {
    throw new Error(`Subscription with key '${key}' already exists`);
  }
  this._subscriptions.push({ key, subscription });
  return this;
}


This would prevent silent overwrites and make debugging easier.

#### 5. Minor: Test organization

The test file has grown to 177 lines. Consider whether the addSubscription tests should be in a separate describe block or even a separate file, especially if more subscription-related features are added in the
future.

### Non-blocking observations

1. Documentation clarity: The README example shows ref("handler", ...) but doesn't explain that this requires the compose framework. Consider adding a note that this is for composed systems.

2. Naming consistency: The method is addSubscription (singular) but returns subscriptions (plural). This is consistent with addAlarm/alarms, so it's fine, but worth noting.

PR #40, opened 2026-04-18
feat(budgets): add BudgetBuilder with recommended thresholds and alarms

laazyj:
## Summary

- Third and final PR derived from #20. Adds `@composurecdk/budgets` — a fluent builder around the `CfnBudget` L1 (the CDK does not ship an L2 for Budgets) with well-architected defaults.
- Surfaces percentage-threshold notification helpers (`notifyOnActual` / `notifyOnForecasted` / `withRecommendedThresholds`), auto-wires `AWS::SNS::TopicPolicy` for any SNS subscriber so `budgets.amazonaws.com` can publish, and offers an opt-in CloudWatch alarm on `AWS/Billing EstimatedCharges`.
- Ships both a combined builder (`createBudgetBuilder`, budget + alarms in one scope) and a standalone alarm builder (`createBudgetAlarmBuilder`, alarms routable into a `us-east-1` stack via `compose().withStacks()`) per [ADR-0004](https://github.com/laazyj/composureCDK/blob/main/docs/adr/0004-split-alarm-builder-for-fixed-region-metrics.md). `addAlarm()` extends both with custom `AlarmDefinitionBuilder<CfnBudget>` definitions.

## Cross-region: `AWS/Billing EstimatedCharges` lives in `us-east-1` only

The `EstimatedCharges` metric is emitted in `us-east-1` only, regardless of where the budget is deployed. CloudWatch alarms are regional, so an alarm in any other region will never fire. Behaviour:

- The combined `createBudgetBuilder` emits a synth-time warning (`@composurecdk/budgets:alarm-region`) when the surrounding stack is not in `us-east-1` — same shape as the existing `cloudfront` and `route53` warnings (suppressed when the region is an unresolved token).
- For non-`us-east-1` deployments, route alarms via `createBudgetAlarmBuilder` into a dedicated `us-east-1` monitoring stack using `compose().withStacks()`. Both builders delegate to a single shared `buildBudgetAlarms()` helper so the recommended-alarm shape and `addAlarm` extension surface stay in lock-step.
- `.budget()` on the standalone builder is **optional** — the recommended `EstimatedCharges` alarm is account-level and does not key off any specific budget. It becomes required only when `addAlarm()` is also called.

The Budget resource itself is a global service and is unaffected by region.

## Review feedback addressed (from #20 / earlier on this branch)

- **`_applyRecommendedThresholds` double-apply** — moved push into `withRecommendedThresholds()` so re-calling `build()` on the same builder doesn't duplicate notifications. Regression test added.
- **`billingAccountId` silently dropped** — removed the field from the public API. The reviewer suggested moving it to `budgetData.accountId`, but `CfnBudget.BudgetDataProperty` has no such field either (verified in the generated CDK types). `AWS::Budgets::Budget` is always created in the deploying account at the CFN level, so the field was never wired up; since the package is unreleased, the cleanest fix is to drop it rather than keep a broken API.
- **Fragile topic-policy dedup key** — switched to `topic.node.path` (fully qualified, unique within the app) for dedup and `topic.node.addr` for the generated construct id suffix. Regression test covers two topics sharing a node id across different scopes.
- **Misleading `thresholdPercent` field name** — renamed to `threshold` on `NotificationEntry` (CFN calls the property `Threshold` and it represents either percent or absolute depending on `thresholdType`). Public helper parameters (`notifyOnActual(thresholdPercent, …)`) keep the percent name since those helpers only emit `PERCENTAGE` thresholds.
- **`us-east-1` constraint should be more strongly enforced and/or sign-posted** — ADR-0004 (which post-dates this PR) defined the project-wide answer for fixed-region-metric services: throw → warn, plus a standalone alarm builder so users can route alarms into `us-east-1` while leaving the resource elsewhere. This PR now follows that pattern, matching the `cloudfront` and `route53` packages.

## Test plan

- [x] `npx nx test budgets` — 51 tests pass (recommended alarm rendering, region warnings on all paths including custom-only, standalone builder + Ref + `compose().withStacks()` cross-region routing, optional `.budget()` semantics).
- [x] `npx nx run-many -t typecheck build test` — all 14 projects clean.
- [x] `npm run lint` clean.
- [x] `npm run format:check` clean.


packages/budgets/src/alarm-config.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Extended {@link AlarmConfig} for the account-level `EstimatedCharges`
+ * billing alarm.
+ *
+ * The `EstimatedCharges` metric lives in the `AWS/Billing` namespace and
+ * is **only emitted in the `us-east-1` region**, regardless of where your
+ * resources run. The alarm created by {@link createBudgetBuilder} will
+ * therefore only receive datapoints when it is synthesised into a stack
+ * deployed to `us-east-1`.
```
laazyj: The `us-east-1` constraint needs to be more strongly enforced and/or sign-posted. Consider if the alarm should _not _ be created if it is being deployed to a different region and/or if the Budget should fail to be created in other regions?

Issue #8, opened 2026-03-29
Real-world consumer


laazyj:
Before releasing v1.0.0 develop a real world application that utilises ComposureCDK to shake out any issues consuming it and demonstrate a minimum coverage of useful CDK components.

laazyj:
Cresting a project to manage resources in my Aws org mgmt account.

This requires support for AWS Organizations, Budgets, and IAM policies and roles

laazyj:
Creating new website to replace my blog. Requires adding Cloudfront functions, Route53, ACM

laazyj:
Route53 is live in jasonduffett.net. CloudFront/ACM is in progress.

laazyj:
Launched at https://jasonduffett.net

PR #47, opened 2026-04-19
feat(cloudfront): add CloudFront Function builder with recommended al…

### LINKED ISSUE #32: Support Cloudfront functions
In PR

laazyj:
…arms

Introduces createFunctionBuilder() alongside the existing distribution builder, since CloudFront Functions are a distinct CDK construct with a custom JS runtime, 1ms compute budget, and no CloudWatch Logs — not just a constrained Lambda. Defaults to the cloudfront-js-2.0 runtime and creates executionErrors / validationErrors / throttles alarms with standard FunctionName + Region=Global dimensions. KeyValueStore is supported as a pass-through prop.

Also extends DistributionBuilder.defaultBehavior.functionAssociations to accept Resolvable<IFunctionRef>, so a function component can be wired to a distribution via ref() within compose(), mirroring the existing origin resolution.

Resolves #32

### DIFF
--- a/packages/cloudfront/src/alarm-config.ts
+++ b/packages/cloudfront/src/alarm-config.ts
@@ -45,3 +45,60 @@ export interface DistributionAlarmConfig {
    */
   originLatency?: AlarmConfig | false;
 }
+
+/**
+ * Controls which recommended alarms are created for a CloudFront Function.
+ * All alarms are enabled by default with sensible thresholds. Set individual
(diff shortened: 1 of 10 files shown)


laazyj:
Exploring an alternative that keeps CloudFront function within the CloudFront DistributionBuilder api

laazyj:
Probably not going to use this. Went with alternative approach in #53 . Will keep open until the alt approach has been proven in #8

PR #50, opened 2026-04-20
feat(route53): add BIND-style zone DSL

laazyj:
## Summary

- Adds `@composurecdk/route53/zone` — a BIND-style DSL for declaring an entire zone as a flat list of records, compiled into the existing per-type builders.
- Covers all eleven record types the package supports: `A`, `AAAA`, `CNAME`, `TXT`, `MX`, `SRV`, `CAA` (plus `CAA_ISSUE`/`CAA_ISSUEWILD`/`CAA_IODEF` wrappers), `NS`, `DS`, `HTTPS`, `SVCB`.
- Records sharing `(type, name)` are merged into one CDK `RecordSet`. Duplicate CNAMEs, apex CNAMEs, and apex NS records are rejected at build time; per-type TTL defaults still apply.
- Ships a production-like example stack (`packages/examples/src/dns-zone-app.ts`) exercising every helper and a new "Zone DSL" section in `packages/route53/README.md`.

## Motivation

The fluent record builders are a good fit for AWS-service records wired to other constructs (CloudFront aliases, ACM validation), but defining a real zone — apex, www, mail, SPF/DMARC/DKIM, CAA, delegation — one builder at a time is clunky. The DSL collapses a whole zone into one declarative list while still resolving through the same builders (defaults, validation, compose wiring via `ref()`).

## Test plan

- [x] `npm run verify` (format, build, lint, test across the monorepo)
- [x] `npx nx test route53` — 54 tests pass, including the new `zone-dsl.test.ts` (merge semantics, APEX handling, CNAME/NS error paths, TTL/comment propagation, compose integration)
- [x] `npx cdk synth ComposureCDK-DnsZoneStack` in the examples package emits `A`, `AAAA`, `CNAME`, `TXT`, `MX`, `SRV`, `CAA`, `NS`, `HTTPS` record sets as expected
- [ ] `npm publish` dry-run of the new subpath export (left for release flow)


packages/route53/src/zone/zone-records.ts
```diff
@@ -0,0 +1,417 @@
...
+    const subScopes = new Map<keyof ZoneRecordsBuilderResult, Construct>();
+    const subScope = (bucket: keyof ZoneRecordsBuilderResult): Construct => {
+      let s = subScopes.get(bucket);
+      if (!s) {
+        s = new Construct(scope, `${id}/${bucket}`);
+        subScopes.set(bucket, s);
+      }
+      return s;
+    };
+    for (const group of groupRecords(this.specs)) {
```
laazyj, line 147: Can this be represented more cleanly with a map/reduce approach?

laazyj:
## Code review

### Overview

Adds `@composurecdk/route53/zone` — a BIND-style DSL that declares an entire zone as a flat list of record helpers and compiles them into the existing per-type builders. `(type, name)` groups are merged into a single CDK record set. Covers all 11 record types the package supports, plus three build-time invariant errors (apex CNAME, duplicate CNAME, apex NS). Ships a subpath export, a new "Zone DSL" README section, and a production-like example stack.

### Strengths

- **Good type modeling.** `RecordSpec` is a proper discriminated union and the switch has an exhaustiveness `_exhaustive: never` branch.
- **Solid test coverage.** 19 tests exercise apex translation, merge semantics, TTL/comment precedence, each record type, compose integration via `ref()`, and the three invariant errors.
- **Defers to existing builders.** Inherits per-type TTL defaults automatically — no risk of drift from the rest of the package.
- **Docs stand on their own.** README helper table + merge semantics + error list + example stack give readers enough to adopt without spelunking.
- **Subpath export is correctly wired.** `package.json` ` ./zone` entry pairs `import` + `types` consistently with the main entry.

### Issues

1. **Construct IDs containing `/` (`zone-records.ts:829`).** ``new Construct(scope, `${id}/${bucket}`)`` — the `constructs` library forbids `/` in IDs (path separator). Worth confirming this actually reaches runtime; if so, switch to a safe separator or nest a parent construct. This is load-bearing for every record emitted by the DSL.
2. **`context` not threaded to inner builders (`zone-records.ts:807` et al.).** `ZoneRecordsBuilder.build` resolves `_zone` with `context`, but inner `.build(scope, id)` calls drop it. Harmless for today's DSL (values are concrete), but inconsistent with every other builder and will silently break if any DSL field becomes `Resolvable` later.
3. **`APEX` vs literal `"apex"` collision (`zone-records.ts:836`).** `const name = head.name === APEX ? "apex" : head.name;` — writing both `A(APEX, …)` and `A("apex", …)` collides on the same construct ID. Reserve a name that can't collide with a valid DNS label.
4. **No validation for empty value arrays.** `A("ha", [])` / `TXT("dmarc", [])` reach CDK with a less-useful error. Cheap to guard in `zoneRecords`.
5. **Silent duplicate values on merge.** `A("www","1.2.3.4")` twice yields two identical addresses in the RR-set. Either de-dupe or call it out in the README merge section.
6. **`applyCommon` structural generic bound.** Works, but a named interface (e.g., `HasCommonRecordOptions`) gives better error messages when a builder is missing one of these methods.
7. **README TTL merge example is ambiguous.** The "first defined ttl/comment wins" semantics of `pickOption` are worth an explicit snippet — easy to misread the current one.

### Risks

- **(1)** is the only potential blocker. Everything else is polish.
- Subpath export freezes public surface once shipped — helper names look well chosen.
- No SPF/DMARC/CAA value validation: intentional, consistent with underlying builders.

### Recommendation

Approve after confirming (1) isn't a runtime error (or swap `/` for `-` defensively). (2) is worth fixing while the surface is new. Also considering adding a shared `constructId()` sanitizer to `@composurecdk/core` so this constraint is enforced uniformly across every package.

laazyj:
## Review feedback addressed + end-to-end deployment verified

Rebased onto #54 (the new \`constructId\` sanitizer in core) and pushed a follow-up commit that addresses every item in the review above.

### Changes since the first review

1. **Construct IDs no longer contain \`/\`** — restructured the tree so \`zoneRecords(...).build(scope, id)\` creates one parent \`Construct\` per instance, one child per bucket (\`a\`, \`aaaa\`, \`cname\`, …), then the record inside. Paths are now \`DNS/records/a/@\` (verified in the deployed stack) instead of silently sanitised to \`DNS--records--a--apex\`.
2. **Context threaded to inner builders** — every \`.build(scope, id, context)\` call now propagates \`context\`.
3. **APEX sentinel used as the key** — the result map and construct ID for the apex group are both the \`@\` sentinel, so \`A(APEX, …)\` and \`A("apex", …)\` no longer collide.
4. **Empty value arrays rejected** — new \`validateSpecs()\` throws a clear error up front instead of letting CDK fail deeper.
5. **Duplicate strings de-duplicated on merge** — identical addresses/values in merged \`A\`/\`AAAA\`/\`TXT\`/\`NS\`/\`DS\` groups are collapsed. Structured values (MX, SRV, CAA, HTTPS, SVCB) are passed through.
6. **\`applyCommon\` bound named** — replaced the inline structural generic with a \`HasCommonRecordOptions<Self>\` interface.
7. **README clarifications** — added an explicit snippet for the first-defined-wins TTL/comment semantics and documented the de-dup behaviour.

Also pulled in a shared \`constructId()\` sanitizer to \`@composurecdk/core\` (#54) so this package — and every future builder — uses one helper for sanitising user-supplied strings into safe construct-ID fragments rather than each rolling its own regex. This PR now depends on #54 and has its base set to \`feat/core-construct-id\`.

### End-to-end verification

Deployed \`ComposureCDK-DnsZoneStack\` to a sandbox account (eu-west-2):

- 19 record sets created (17 user records + Route 53's auto apex NS/SOA).
- Every group that should merge did so in one RR-set: \`A("api", …) × 2\` → one record with both IPs; 2 × \`MX\`, 2 × \`SRV\`, 4 × \`CAA\` (\`ISSUE\` × 2 + \`ISSUEWILD\` + \`IODEF\`) each produced a single record.
- TTL precedence worked as documented: \`api\` inherited TTL 60 from the second call because the first had none.
- Construct paths in the CloudFormation event log were clean — no \`--\` sanitisation anywhere (\`DNS/records/a/@\`, \`DNS/records/srv/_sip._tcp\`, etc.).
- Stack output \`NameServers\` resolved correctly via the \`outputs()\` hook + \`ref().get("hostedZone")\`.

Stack was destroyed after verification.

### Example stack: zone name changed

The demo zone in \`packages/examples/src/dns-zone-app.ts\` now uses \`composurecdk-zone-dsl-demo.com\` instead of \`example.com\`. Route 53 refuses to create a hosted zone for \`example.com\` (IANA-reserved) — the domain is still fine as a record *value*, so the record targets elsewhere in the example are unchanged.

PR #75, opened 2026-05-03
feat(budgets)!: encode service constraints in NotifySubscribers + Email

laazyj:
## Summary

- Replace variadic `BudgetSubscriber[]` with a structured `NotifySubscribers { sns?, emails? }` shape — passing two SNS topics is now unrepresentable in TypeScript.
- Add branded `Email` type with a validating `email()` factory so bare strings are rejected at compile time.
- Validate `BudgetLimit.unit` and `EstimatedChargesAlarmConfig.currency` against `DEFAULT_BUDGET_CURRENCIES` (hard error on unknown ISO 4217 strings, soft warning on non-USD).
- Cap email subscribers at 10 per notification at synth time (AWS allows 1 SNS + 10 EMAIL).

## Why

A user hit a 5-minute partial-stack rollback from this code:

```ts
.notifyOnActual(100, alertTopic, killSwitchTopic)  // two SNS topics
```

CloudFormation accepted the template; the Budgets service rejected the request mid-deploy with `Unable to create subscriber - one notification can only have 1 subscribers with type of SNS`. A prior bug had the same shape (`Unit: "GBP"` accepted by CFN, rejected at deploy by a USD-only billing account).

Both are "service rejects what CFN accepts" — the kind of constraint the builder should encode rather than punt to a slow round-trip. This change moves both into the type system / synth-time validation.

## Migration

```ts
// before
.notifyOnActual(100, alertTopic, "ops@example.com")
.withRecommendedThresholds("ops@example.com")

// after
import { email } from "@composurecdk/budgets";

.notifyOnActual(100, { sns: alertTopic, emails: [email("ops@example.com")] })
.withRecommendedThresholds({ emails: [email("ops@example.com")] })
```

The `BudgetSubscriber` type is removed from the public API. `NotificationEntry.subscribers` is retyped to `NotifySubscribers` so the escape hatch carries the same constraint.

## Test plan

- [x] `npx nx test budgets` — 67 tests pass (added: email factory, >10-email cap, 1+10 max-allowed, ZZZ rejection, GBP soft-warning, USD no-warning, USAGE skip, type-level bare-string rejection, 1 SNS + N EMAIL on a single notification mirroring the original bug)
- [x] `npx nx run-many -t build typecheck test` — green across 15 projects
- [x] `npm run lint` and `npm run format:check` — clean
- [ ] Verify the migrated call site for the user's original bug synthesises with 1 SNS + 1 EMAIL on the 100% notification (covered by the new "original bug's fix" test case)
- [ ] Confirm `unit: "ZZZ"` errors at synth and `unit: "GBP"` annotates with `@composurecdk/budgets:limit-currency`

### DIFF
--- a/packages/budgets/README.md
+++ b/packages/budgets/README.md
@@ -7,12 +7,12 @@ This package provides a fluent builder for `AWS::Budgets::Budget` with well-arch
 ## Budget Builder
 
 ```ts
-import { createBudgetBuilder } from "@composurecdk/budgets";
+import { createBudgetBuilder, email } from "@composurecdk/budgets";
 
 const budget = createBudgetBuilder()
   .budgetName("AgentBudget")
-  .limit({ amount: 50, unit: "GBP" })
-  .notifyOnActual(100, "ops@example.com")
+  .limit({ amount: 50 })
+  .notifyOnActual(100, { emails: [email("ops@example.com")] })
   .build(stack, "AgentBudget");
 ```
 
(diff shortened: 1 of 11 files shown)


laazyj (commented):
Look good

PR #77, opened 2026-05-05
feat: builder-level .tag() / .tags() via decorator wrapper + tags() afterBuild hook

laazyj:
## Summary

Closes #66.

Adds builder-level `.tag(key, value)` / `.tags({...})` to every builder in the library via a shared decorator wrapper, plus a complementary `tags()` afterBuild hook for system-wide cross-cutting tags.

The mechanism is a decorator factory `taggedBuilder<Props, T>(constructor)` that wraps `core.Builder()`. Every builder factory in the library opts in by calling `taggedBuilder<>()` instead of the bare `Builder<>()`. `@composurecdk/core` is unchanged — the decorator and its tag-application logic live in `@composurecdk/cloudformation` alongside `StackBuilder` and `outputs()`. A new `no-restricted-imports` ESLint rule under `packages/*/src/**` enforces that future builders pick up the decorator by default.

The architectural decision being recorded is the **decorator-builder pattern itself** (ADR-0005), not the tagging feature — the pattern is the structural primitive for layering future cross-cutting features (telemetry, dry-run, etc.) onto every builder without modifying `core` or duplicating mechanics.

## Commits

1. **`feat(cloudformation): add taggedBuilder wrapper, tag validator, and result-walker`** — the decorator factory, the `applyBuilderTags` walker, the `validateTag` validator, and 24 unit tests covering type augmentation, validation, override warnings, walker semantics (constructs, Records, CDK-core-non-constructs), and `Tags.of` equivalence.
2. **`feat(s3)!: wire BucketBuilder to taggedBuilder`** — first concrete builder to migrate. BucketBuilder is the proof-of-concept because its result exercises every walker path: primary bucket, optional sibling `accessLogsBucket`, and an `alarms: Record<string, Alarm>` map.
3. **`refactor(cloudformation): route StackBuilder through taggedBuilder`** — behaviour-preserving migration of the only existing builder with its own bespoke `.tag()`. `toScopeFactory()` reads accumulated tags via the wrapper's `getBuilderTags(this)` escape hatch.
4. **`feat!: roll taggedBuilder out across all remaining builders + lint enforcement`** — mechanical migration of every other builder factory (acm, apigateway, budgets, cloudfront, ec2, iam, lambda, logs, route53, sns), peer-dep additions for `@composurecdk/cloudformation`, and the `no-restricted-imports` lint rule.
5. **`feat(cloudformation): add tags() afterBuild helper, ADR for decorator builder pattern, docs, example`** — Layer 2 hook (`tags({ system, byComponent })`), ADR-0005 documenting the decorator-builder pattern, `docs/extensions.md` updates, and `packages/examples/src/tagged-system-app.ts` demonstrating both layers against a downstream selector-tag use case.

## Behaviour summary

**Layer 1 — `.tag()` / `.tags()` on every builder.** Tags accumulate during configuration (insertion-ordered map, last-wins on duplicate, with `process.emitWarning` so overrides are visible). At `build()`, the walker tags every top-level `IConstruct` field in the result and every value inside top-level `Record<string, IConstruct>` fields. Wrapper objects, primitives, arrays, and CDK core non-constructs (`PolicyDocument`) are skipped.

**Layer 2 — `tags()` afterBuild hook.** `system` entries reach every taggable construct via `Tags.of(scope).add(...)` on the top-level scope. `byComponent` entries reach a specific component's scope (its per-component stack under `withStacks`/`withStackStrategy`). Component keys are statically typed against the composed system's keys.

**Validation.** Both layers validate keys/values eagerly: rejects empty keys, the reserved `aws:` prefix (case-insensitive), keys >128 chars, values >256 chars, and characters outside the AWS-documented tag character set.

**Precedence.** Builder-level tags win on key collision because they target a closer scope — CDK's tag priority resolves automatically.

## Breaking changes

- Every `IXxxBuilder` alias structurally extends `ITaggedBuilder<Props, T>` instead of `IBuilder<Props, T>`. Pure consumers that destructure or annotate against the old shape need to recompile.
- `StackBuilder.tag()`'s duplicate-key behaviour changes: previously appended silently (last-wins observable in CFN tags), now last-wins with `process.emitWarning`. Functionally compatible; previously-quiet duplicate calls now emit a warning.
- Every builder package picks up `@composurecdk/cloudformation` as a peer dependency.

## Open questions for review

I addressed every open question raised in the issue thread, but a few remain in scope for follow-up review:

1. **`DistributionBuilderResult.functions` is `Record<string, FunctionEntry>`** — the wrapped `CfFunction` constructs inside `FunctionEntry` are not reached by the one-level-deep walker. ADR-0005 acknowledges this; the documented options are (a) flatten the result, or (b) extend the walker with a documented escape hatch. I have not done either yet — preserved current behaviour, which is "the CfFunction inherits stack-level tags via Layer 2 propagation but doesn't get builder-level tags." Picking (a) or (b) is a follow-up depending on real-world impact.
2. **`*BuilderResult` interfaces include constructs the user might not expect to be tagged.** The walker tags everything in the result. If a future builder wants to expose a construct in the result without tagging it, an opt-out marker (e.g. a symbol-keyed `__skipTagging` field) would be needed. None of the existing builders need this. Deferring until a real case appears.
3. **Stacked decorators.** ADR-0005 documents that decorators compose by stacking proxies, with the outermost factory determining the type the user sees. Today only `taggedBuilder` exists; the pattern is forward-compatible for additional decorators but I haven't proven the stacking ergonomics with a real second decorator yet. This will be exercised the first time a second cross-cutting feature lands.
4. **Existing tests in `packages/cloudformation/test/stack-builder.test.ts`** don't assert that `toScopeFactory()` snapshots tags at call time and ignores later mutations. The invariant is preserved by construction (a code reviewer confirmed) but isn't test-locked. Pre-existing gap, not introduced here — flagging in case it's worth a quick test on the way through.

## Test plan

- [ ] `npm run lint` passes
- [ ] `npm run format:check` passes
- [ ] `npx nx run-many -t build,typecheck,test` — 15 projects, all green (62 cloudformation tests, 76 s3, 54 lambda, 61 examples, plus the rest)
- [ ] ESLint rule fires when a builder file imports `Builder` from `@composurecdk/core` (verified locally with a smoke test)
- [ ] `packages/examples/src/tagged-system-app.ts` synthesises and the test asserts (a) selector tag lands only on the EC2 instance, (b) system-wide tags reach the bucket, VPC, instance, and flow-log LogGroup
- [ ] Existing snapshot tests for the other example apps still match (no incidental tag changes leaked into them)

### DIFF
--- a/docs/adr/0005-decorator-builder-pattern.md
+++ b/docs/adr/0005-decorator-builder-pattern.md
@@ -1,0 +1,63 @@
+# ADR 0005: Decorator pattern for adding cross-cutting builder features
+
+- **Status:** Accepted
+- **Date:** 2026-05-05
+
+## Context
+
+`@composurecdk/core` exports `Builder<Props, T>(constructor)` and the `IBuilder<Props, T>` mapped type. Together they provide a deliberately minimal, CDK-agnostic builder primitive: a Proxy that intercepts prop setters and chainable methods, with no opinion about what the wrapped class does. Every builder package in the library — `s3`, `lambda`, `apigateway`, `ec2`, `route53`, etc. — uses this primitive directly to build a fluent surface around a CDK construct.
+
+Some features are intrinsically cross-cutting: they apply to **every** builder rather than to one resource type. Examples that have come up or are likely to:
+
+- **Tagging** — every taggable construct should be reachable from a single `.tag(key, value)` / `.tags({...})` call on the builder, with the tag landing on the primary construct and on every sibling the builder creates (auto-managed log groups, access-log buckets, alarms maps).
+- **Future candidates** — uniform `.timeout()` / `.retries()` for build-time IAM wiring waits, structured logging or telemetry hooks at `build()` time, a `.dryRun()` switch that returns a result without committing CDK constructs, removal-policy defaults, etc.
+
+When tagging — the first cross-cutting feature — was being designed (issue #66), three implementation paths surfaced:
+
+1. **Per-builder methods.** Each of ~30 builder classes declares its own `.tag()`, its own `#tags` field, and its own application call in `build()`. Roughly 10 lines × 30 builders = ~300 lines of duplicated mechanics. High drift risk: any new builder must remember to add the methods, and any change to the validator or accumulator semantics must be applied in every package.
+2. **Modify `core.Builder()`** to special-case `.tag()` and `.tags()` (and any future cross-cutting method). Eliminates duplication, but couples `core` to `aws-cdk-lib`'s `Tags` API and bloats the proxy with feature-specific name handling. The boundary `core` keeps — CDK-agnostic, builder primitive only — was a deliberate decision and we shouldn't relax it for one feature, let alone for an open-ended list of future features.
+3. **A decorator factory that wraps `Builder()`.** A new factory in a CDK-aware package returns a decorated proxy that intercepts the cross-cutting method names before delegating to the inner core proxy, applies behaviour at `build()` time, and re-wraps chainable returns so the decorator surface stays reachable.
+
+Path (3) is the standard decorator/composite pattern: the decorator implements the same interface as the base, adds behaviour, and forwards everything else. CDK-side TypeScript developers will recognise it from the way Aspects layer onto constructs without changing them, and from the way `@composurecdk/cloudformation` already layers `outputs()` and `StackBuilder` on top of `core`.
+
+## Decision
+
+**Cross-cutting builder features are added via decorator factories that wrap `Builder()` from `@composurecdk/core`. Each builder factory in the library opts into the decoration by calling the decorator instead of the bare core API. `core` itself stays minimal and CDK-agnostic.**
+
+Concrete shape:
+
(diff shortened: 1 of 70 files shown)


laazyj:
Pushed `8ef94d7` for issue #3.

**Approach**
After looking at the trade-offs (parallel maps vs walker descent vs opt-in marker), went with generalising the walker. The result-type API is unchanged; `FunctionEntry` stays as it is. The walker now recursively descends through plain-object literals and tags every `IConstruct` it finds. CDK's tag aspect handles propagation across each construct's subtree.

**Boundaries**
- Constructs are tagged via `Tags.of(...).add(...)`; recursion stops there.
- Class instances that aren't constructs (PolicyDocument etc.) remain opaque — `isPlainObject` requires `Object.prototype` as the prototype, so recursion stops at any class boundary.
- Arrays and primitives are skipped.

**Why not split into two parallel maps**
Considered `functions: Record<string, CfFunction>` + `functionMeta: Record<string, FunctionMetadata>`, but the duplication felt like API gymnastics to work around a walker limitation. The deeper-walker approach implements the contract that's already implicit in `architecture.md`: every construct exposed in a builder's result type is a tag target. Wrapper shapes like `Record<string, { construct, metadata }>` are unwrapped naturally; authors don't need an opt-in marker.

**Tests**
- New test asserts both default-behavior and additional-behavior CloudFront Functions receive `.tag()`/`.tags()` entries.
- The "wrappers" test in `apply-builder-tags.test.ts` was previously asserting no-recursion; flipped to assert the new behaviour.
- Existing `Tags.of equivalence` and `does nothing when no tags accumulated` tests still pass.

Issues 4–10 are smaller; happy to pick them up next session.

PR #147, opened 2026-05-25
feat(cloudwatch): accept MathExpression in AlarmDefinition metric

laazyj:
## Summary

Widens the alarm metric type so rate/ratio alarms expressed as `MathExpression` (e.g. `errors / invocations`) compile against `AlarmDefinition` and the fluent `AlarmDefinitionBuilder` — no casting or vendoring required.

- New exported convenience type `AlarmMetric = Metric | MathExpression`, used by both `AlarmDefinition.metric` and `AlarmDefinitionBuilder.metric()`.
- The union targets the **concrete classes**, not `IMetric`: `createAlarm` is declared on `Metric`/`MathExpression`, not on the `IMetric` interface, and both share an identical `createAlarm` signature — so `createAlarms` needs no change.
- **Pure type widening, no runtime/behaviour change.** Existing recommended-alarm definitions all return `Metric`, which is assignable to the union, so the ~15 resource packages are untouched.

Closes #144.

## MathExpression behaviour (documented, not enforced in code)

Verified against `aws-cdk-lib@2.254.0`:

- **Period lives on the expression.** `MathExpression` overrides the period of every metric in `usingMetrics` to its own `period` (default 5 min) and warns when they differ — so the README example sets `period` on the expression.
- **Ratio/divide-by-zero** drops the data point and can flip a ratio alarm to `INSUFFICIENT_DATA`; the example uses `IF(invocations > 0, errors / invocations, 0)` and notes `treatMissingData`.
- **Single time series only.** `SearchExpression` is a separate class (excluded by the union for free); a `MathExpression` using `SEARCH(...)` that returns multiple series is rejected by CloudWatch at deploy — documented.

## Changes

- `src/alarm-definition.ts` — add `AlarmMetric`, widen `metric` field.
- `src/alarm-definition-builder.ts` — widen `#metricFactory` and `metric()`.
- `src/index.ts` — export `AlarmMetric`.
- `README.md` — document the union and add a rate-alarm example.

## Tests

- `test/create-alarms.test.ts` — alarm from a `MathExpression` synthesizes a `Metrics` array (with top-level `MetricName`/`Statistic` absent).
- `test/alarm-definition-builder.test.ts` — the builder factory accepts and resolves a `MathExpression`.

## Verification

`npm run lint`, `npm run format:check`, `npx nx test cloudwatch` (82 passing), and the full `npm run verify` gate (build + `check:exports` attw/publint confirming the new `AlarmMetric` export emits for both ESM and CJS) all pass.

### DIFF
--- a/packages/cloudwatch/README.md
+++ b/packages/cloudwatch/README.md
@@ -46,6 +46,40 @@ The `description` method also accepts a factory for contextual descriptions:
 builder.description((def) => `Alert when invocations >= ${def.threshold} per minute`);
 ```
 
+### Rate and ratio alarms
+
+The metric factory may return either a `Metric` or a `MathExpression` (the
+exported `AlarmMetric` union), so rate/ratio alarms share the same definition
+shape:
+
+```ts
+import { MathExpression } from "aws-cdk-lib/aws-cloudwatch";
+
+new AlarmDefinitionBuilder<LambdaFunction>("errorRate")
+  // Guard the denominator: a divide-by-zero data point is dropped, which would
+  // otherwise push the alarm into INSUFFICIENT_DATA when there are no invocations.
+  .metric(
+    (fn) =>
+      new MathExpression({
+        expression: "IF(invocations > 0, errors / invocations, 0)",
+        usingMetrics: {
+          errors: fn.metricErrors(),
+          invocations: fn.metricInvocations(),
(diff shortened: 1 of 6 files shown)


laazyj (commented):
Looks good. Ship it.

PR #151, opened 2026-05-25
feat(eslint-plugin): ban CDK APIs above the floor

laazyj:
> **Stacked on #148** (base is `fix/cloudwatch-alarm-iscfn-compat`). The diff shown is only the new rule; it retargets to `main` once #148 merges. Depends on #148 because the rule would otherwise flag the pre-fix cloudwatch code on `main`.

Step 1 of the cross-version compatibility plan (follow-up to #146 / #148).

## What

A new internal lint rule, `composurecdk/no-cdk-api-above-floor`, that flags use of `aws-cdk-lib` APIs newer than the supported peer-dependency floor. These compile fine — devDeps track the latest CDK — but throw at runtime for consumers on an older, still-supported version. That's exactly how #146 shipped undetected.

Seeded with the per-resource `isCfn<Resource>` L1 static type guards (e.g. `CfnAlarm.isCfnAlarm`), which aws-cdk-lib first shipped in **2.231.0** (verified by installing real versions — 2.230.0 lacks them, 2.231.0 has them):

- Matched by **member name** (`/^isCfn[A-Z]/`), so it holds regardless of how the class is imported/aliased.
- Allowlists the foundational `isCfnResource` / `isCfnElement` (core since v2.0 — `isCfnResource` is in fact the portable replacement).
- Only fires on **member access** (`Class.method`), so the bare-call helpers we ship (`isCfnAlarm(node)`) are unaffected.

The `FORBIDDEN` list is deliberately a growable array — more entries land as we pin and lower the floor toward 1.0.0 (steps 2–3).

## Why a lint rule

A static rule prevents the call from ever being *written* in `src/`, in-editor, before any build — strictly better than a runtime check for "don't call this". The complementary runtime guarantee (synth against a *real* old CDK) is the next PR.

## Tests / checks

- `RuleTester` cases: bare-call helper, `CfnResource.isCfnResource`, `CFN_RESOURCE_TYPE_NAME`, and a helper *definition* are all valid; `CfnAlarm.isCfnAlarm`, `CfnCompositeAlarm.isCfnCompositeAlarm`, and `CfnBucket.isCfnBucket` are flagged.
- Plugin build + tests green; **repo-wide `npm run lint` passes with the rule active** (confirms no `src/` still reaches for a version-gated API after #148); format clean.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-cdk-api-above-floor.ts
+++ b/packages/eslint-plugin/src/rules/no-cdk-api-above-floor.ts
@@ -1,0 +1,89 @@
+import type { Rule } from "eslint";
+import type { MemberExpression } from "estree";
+
+/**
+ * An `aws-cdk-lib` API that postdates @composurecdk's supported peer-dependency
+ * floor, and so must not be used in library `src/`: calling it throws on older
+ * (but still supported) CDK versions.
+ *
+ * Matched by the accessed member name rather than the owning class identifier,
+ * so it holds up regardless of how the class is imported or aliased.
+ */
+interface ForbiddenMember {
+  /** True when an accessed property name is the banned member. */
+  matches: (propertyName: string) => boolean;
+  /** Short description of the banned API for the diagnostic. */
+  label: string;
+  /** The `aws-cdk-lib` version that introduced it. */
+  since: string;
+  /** What to use instead. */
+  use: string;
+}
+
+/**
+ * Foundational `isCfn*` guards that predate the floor (core, since v2.0) and
+ * are therefore allowed — `isCfnResource` is in fact the portable replacement
+ * for the banned per-resource guards.
+ */
+const ALLOWED_CFN_GUARDS = new Set(["isCfnResource", "isCfnElement"]);
+
+/**
+ * The ban list. Add an entry whenever an `aws-cdk-lib` API we reach for turns
+ * out to postdate the supported floor — this list is expected to grow as the
+ * floor is pinned and lowered ahead of 1.0.0.
+ *
+ * Seeded with the per-resource `isCfn<Resource>` static type guards, which
+ * aws-cdk-lib first shipped on every generated L1 in 2.231.0 (verified by
+ * installing real versions: 2.230.0 lacks them, 2.231.0 has them). Calling
+ * them (e.g. `CfnAlarm.isCfnAlarm(node)`) throws `TypeError` on the older
+ * versions in our `^2` peer range — see issue #146.
+ */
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin/src/rules/no-cdk-api-above-floor.ts
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * Seeded with the per-resource `isCfn<Resource>` static type guards, which
+ * aws-cdk-lib first shipped on every generated L1 in 2.231.0 (verified by
+ * installing real versions: 2.230.0 lacks them, 2.231.0 has them). Calling
+ * them (e.g. `CfnAlarm.isCfnAlarm(node)`) throws `TypeError` on the older
+ * versions in our `^2` peer range — see issue #146.
+ */
+const FORBIDDEN: ForbiddenMember[] = [
+  {
+    matches: (name) => /^isCfn[A-Z]/.test(name) && !ALLOWED_CFN_GUARDS.has(name),
```
laazyj, line 55: This seems dangerously greedy. Any `isCfn*` function name would trigger this right? It doesn't appear to be limited to `aws-cdk-lib/*` members.

laazyj: Good catch — fixed in 46ccc87. Two layers:

1. **`chainRoot`** walks the member chain to its leftmost identifier through `MemberExpression`, `ChainExpression`, and the TS wrappers (`as`, `!`, `satisfies`, angle-bracket assertion). So `(CfnAlarm as any).isCfnAlarm(x)` can't smuggle past, and a call in the chain (`Stack.of(x).isCfnY()`) still breaks the link as before.
2. **Scope-aware resolution** — the root identifier is resolved through `ctx.sourceCode.getScope(node)`, so the rule fires only when it binds to an `aws-cdk-lib*` `ImportDeclaration`. A local that shadows the import name (`function f(CfnAlarm) { CfnAlarm.isCfnAlarm(x) }`) correctly resolves to the inner binding and is not flagged.

Reviewed twice more after the initial fix (each pass surfaced one further real issue — TS-wrapper bypass, then a runtime crash on `import = require` because typescript-eslint reports that as `ImportBinding` with no `.source`); both folded in. Tests cover all idiomatic CDK import styles (named, aliased, `* as` submodule, named submodule as namespace, deep `cdk.aws_*.X`, optional chain), TS wrappers, shadowing, allow-listed members on a cdk root, calls that break the chain, and the `import = require` no-crash guard — 51 cases. Documented gaps: computed bracket access and `import = require` are silently untracked (uncommon in ESM src).

PR #155, opened 2026-05-25
test(ci): aws-cdk-lib floor compatibility harness

laazyj:
> **Stacked on #148** (base is `fix/cloudwatch-alarm-iscfn-compat`). Diff shown is only the harness; retargets to `main` once #148 merges. Depends on #148 — the synth would fail against the *unfixed* package, which is the point.

Steps 2–3 of the cross-version compatibility plan (follow-up to #146 / #148 / #151). A real-install harness — no simulation — in two complementary halves.

## 1. Composed-synth smoke (always-on PR gate)

`scripts/cdk-floor-test.mjs` → the `cdk-floor` job in `ci.yml`, and `npm run test:cdk-floor`:

- Discovers and `npm pack`s **every publishable `@composurecdk` package** (16).
- Installs them into a throwaway project with a **real pinned `aws-cdk-lib`** + constructs.
- Synthesises a representative `compose(...).build()` system — core + sqs + lambda + sns + cloudwatch (and their peers: cloudformation, iam, logs) — and asserts **invariants** (resource counts, every alarm wired by `alarmActionsPolicy`), never exact CFN output. Drift-robust, so it's a reliable gate.

Verified locally on real `aws-cdk-lib@2.230.0`: 11 alarms, all wired, ✓.

## 2. Suites-at-floor (manual floor-finding)

`scripts/cdk-floor-suites.mjs` → the `cdk-floor-suites` **`workflow_dispatch`** workflow (takes a version input), and `npm run test:cdk-floor-suites`:

- Pins the chosen `aws-cdk-lib` across the workspace and runs **every package's own unit suite** (partial matchers — drift-robust) + an example-stack synth (`test:update`, the broadest integration surface).
- **Not a PR gate** — exact-output assertions drift across versions, so a failure means "investigate", not "broken". It mutates the workspace, so it's gated behind `CI`/`--force`.

## Why both

The static lint rule (#151) stops a too-new API being *written*; the composed smoke proves the published packages *synth* on the floor; the suites sweep finds *where* the floor actually is, per package.

## Moving the floor (step 3)

One `CDK_FLOOR` constant / env var drives both, e.g. `CDK_FLOOR=2.46.0 npm run test:cdk-floor`. Empirically so far: cloudwatch alone synths down to **2.1.0**; the composed system's floor is gated by the newest CDK feature any builder uses (e.g. the example processor's Lambda runtime enum). Lowering `CDK_FLOOR` and re-running the suites workflow is how we'll pin the real supported floor — which should then drive `peerDependencies` bumps + an ADR before 1.0.0.

### DIFF
--- a/.github/workflows/cdk-floor-suites.yml
+++ b/.github/workflows/cdk-floor-suites.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: 20
+          node-version: 24
           cache: npm
 
       - name: Install dependencies
--- a/.github/workflows/cdk-floor-suites.yml
+++ b/.github/workflows/cdk-floor-suites.yml
@@ -1,0 +1,44 @@
+name: CDK floor suites
+
+# Floor-finding tool: runs every package's unit suite against a pinned
+# aws-cdk-lib version to discover the lowest the library still supports. Run it
+# by hand from the Actions tab with a version; lower the version and re-run as
+# you narrow the floor. Not a PR gate — the always-on gate is the drift-robust
+# composed synth in the `cdk-floor` job of ci.yml.
+
+on:
+  workflow_dispatch:
+    inputs:
+      cdk-version:
+        description: aws-cdk-lib version to test the suites against
+        required: true
+        default: 2.230.0
+
+permissions:
+  contents: read
+
+jobs:
+  suites:
+    name: Suites against aws-cdk-lib ${{ inputs.cdk-version }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    env:
+      NX_DAEMON: false
+      CDK_FLOOR: ${{ inputs.cdk-version }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 1
+
+      - name: Setup Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 20
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run the unit suites against the floor
+        run: npm run test:cdk-floor-suites
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: 20
+          node-version: 24
           cache: npm
 
       - name: Install dependencies
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,3 +60,33 @@ jobs:
 
       - name: Test
         run: npm run test
+
+  cdk-floor:
+    name: CDK floor compatibility
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    env:
+      NX_DAEMON: false
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 1
+
+      - name: Setup Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 20
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      # Packs from each package's dist, so the build must run first.
+      - name: Build
+        run: npm run build
+
+      # Installs a real, pinned aws-cdk-lib floor and synthesises against it,
+      # catching published packages that reach for a too-new CDK API (#146).
+      - name: Synth against the aws-cdk-lib floor
+        run: npm run test:cdk-floor
--- a/scripts/cdk-floor-suites.mjs
+++ b/scripts/cdk-floor-suites.mjs
@@ -1,0 +1,67 @@
+#!/usr/bin/env node
+
+/**
+ * aws-cdk-lib floor compatibility harness (exhaustive-suites half).
+ *
+ * Pins a real aws-cdk-lib floor across the workspace and runs every package's
+ * own unit suite against it, so a too-new CDK API in *any* package surfaces —
+ * not just the composed smoke that `cdk-floor-test.mjs` exercises.
+ *
+ * This is a floor-*finding* tool, run manually (the `cdk-floor-suites` workflow,
+ * dispatched with a version). It is NOT a PR gate: the per-package suites use
+ * partial matchers and mostly hold across versions, but exact-output assertions
+ * drift, so a failure means "investigate", not necessarily "broken". The
+ * always-on PR gate is the drift-robust composed synth in `cdk-floor-test.mjs`.
+ *
+ * MUTATES the workspace (installs the floor over node_modules; the example
+ * suites rewrite their snapshots). Intended for ephemeral CI — run `npm ci`
+ * (and `git checkout packages/examples/test/__snapshots__`) to restore locally.
+ */
+
+import { execFileSync } from "node:child_process";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const REPO_ROOT = resolve(dirname(fileURLToPath(import.meta.url)), "..");
+const FLOOR = process.env.CDK_FLOOR ?? "2.230.0";
+
+if (process.env.CI === undefined && !process.argv.includes("--force")) {
+  console.error(
+    "cdk-floor-suites mutates node_modules and rewrites example snapshots — it is meant for CI.\n" +
+      "Run with --force to proceed locally, then restore with `npm ci` and " +
+      "`git checkout packages/examples/test/__snapshots__`.",
+  );
+  process.exit(1);
+}
+
+function run(cmd, args, extraEnv) {
+  execFileSync(cmd, args, {
+    cwd: REPO_ROOT,
+    stdio: "inherit",
+    env: extraEnv ? { ...process.env, ...extraEnv } : process.env,
+  });
+}
+
+console.log(`Pinning aws-cdk-lib@${FLOOR} across the workspace …`);
+run("npm", ["install", "--no-save", "--no-audit", "--no-fund", `aws-cdk-lib@${FLOOR}`]);
+
+// Builder/core suites use partial matchers, so they assert against the floor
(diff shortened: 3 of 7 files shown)


.github/workflows/cdk-floor-suites.yml
```diff
@@ -0,0 +1,44 @@
+name: CDK floor suites
```
laazyj, line 1: I understand that running the floor test script is more ergonomic to do through CI as it doesn't mess up a durable workspace. 

How is the output of the run displayed and analysed by the user?

.github/workflows/cdk-floor-suites.yml
```diff
@@ -0,0 +1,44 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 1
+
+      - name: Setup Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 20
```
laazyj: Why such an old version of Node? If no specific reason prefer 24 to defer EOL upgrades

.github/workflows/ci.yml
```diff
@@ -60,3 +60,33 @@ jobs:
 
       - name: Test
         run: npm run test
+
+  cdk-floor:
```
laazyj, line 64: **Question**: Should we consider the cdk-floor test as a separate action rather than bundled in ci.yml? There may be benefits in treating it as a distinct workflow/action compared to the base lint/build/test action.

Are there benefits (i.e. permissions) that are simpler in a single `ci.yml` file?

scripts/cdk-floor/synth.mjs
```diff
@@ -0,0 +1,76 @@
...
+const version = require("aws-cdk-lib/package.json").version;
+console.log(`aws-cdk-lib ${version}`);
+
+const app = new App();
+const stack = new Stack(app, "FloorStack");
+
+// A queue → Lambda consumer plus an SNS alert topic, with all alarms routed to
+// the topic. Exercises core (compose/ref), sqs, lambda (event source + role),
+// sns, and cloudwatch (recommended + custom alarms, actions policy) — and their
+// peers (cloudformation, iam, logs) transitively.
```
laazyj, line 34: **Question**: How will this smoke test scale as we support more and more AWS resource types? Already it fails to exercise `acm`, `route53`, and `ec2` packages. Does this matter, or is the "smoke test" still proving something valuable with incomplete coverage?

Similarly, it tests the `alarmActionsPolicy`, but no other policies? Is this because that policy is special or an exemplar that proves coverage beyond its scope? Or is it just copied from other examples?

scripts/cdk-floor/synth.mjs
```diff
@@ -0,0 +1,76 @@
...
+const alarms = Object.values(template.findResources("AWS::CloudWatch::Alarm"));
+if (alarms.length === 0) throw new Error("expected the composed system to produce alarms");
+const unwired = alarms.filter((a) => !Array.isArray(a.Properties.AlarmActions));
+if (unwired.length > 0) {
+  throw new Error(`alarmActionsPolicy left ${unwired.length}/${alarms.length} alarm(s) unwired`);
+}
+
+console.log(
+  `PASS: composed system synthesised on aws-cdk-lib ${version} ` +
+    `(${alarms.length} alarms, all wired to the alert topic)`,
```
laazyj, line 75: What's so special about alarms that is a significant assertion of the test?

.github/workflows/cdk-floor-suites.yml
```diff
@@ -0,0 +1,44 @@
...
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run the unit suites against the floor
+        run: npm run test:cdk-floor-suites
```
laazyj, line 44: How is the `CDK_FLOOR` version intended to be passed to the CI workflow? Is this a set once/occasionally env var? Or a per-run input variable for  the action?

scripts/cdk-floor-suites.mjs
```diff
@@ -0,0 +1,67 @@
...
+function run(cmd, args, extraEnv) {
+  execFileSync(cmd, args, {
+    cwd: REPO_ROOT,
+    stdio: "inherit",
+    env: extraEnv ? { ...process.env, ...extraEnv } : process.env,
+  });
+}
+
+console.log(`Pinning aws-cdk-lib@${FLOOR} across the workspace …`);
+run("npm", ["install", "--no-save", "--no-audit", "--no-fund", `aws-cdk-lib@${FLOOR}`]);
```
laazyj, line 46: How will this script work if the individual packages support different floor versions?

laazyj (commented):
This seems like it doesn't quite meet the goals of the test. Perhaps it is only an intermediate step toward where we want to be. Questions left inline.

.github/workflows/cdk-floor-suites.yml
laazyj: Honest answer: it streams to the workflow log — the `nx run-many` lines (pass/fail per project), then the example synths, ending with the final `✓`/`✗`. A real failure surfaces as `Named export 'X' not found from 'aws-cdk-lib/aws-Y'`, with the package and missing export named, so it's actionable but you have to scroll.

That's not a great UX for floor-finding. A structured summary — `$GITHUB_STEP_SUMMARY` table or a per-package JSON output — would be much better and is easy to add if we keep this tool. But see the main review thread: with per-package floors landing in #156, this script's role narrows considerably, and the proper enforcement signal lives there (`cdk-floors:enforce` fails with a clear per-package list).

.github/workflows/cdk-floor-suites.yml
laazyj: Bumped to Node 24 in b70a879 (and the same Node 20 pin in the `cdk-floor` job in `ci.yml` — both were mirroring the main job's floor without a real reason).

.github/workflows/ci.yml
laazyj: Folded into `ci.yml` because: same triggers (every PR), same checkout/`setup-node`/`npm ci` setup, and no permission difference (`contents: read` only) — splitting would just duplicate the scaffolding. The split *would* help if we wanted independent concurrency, different cadence (e.g. cron), or to gate it differently (label-only, paths-filter, etc.). None of those needs exist yet; trivial to split later if any does.

The `cdk-floor-suites` workflow IS already separate, because its trigger is `workflow_dispatch` (manual ad-hoc version input) — that *is* a different shape from PR-driven CI and warrants its own file.

scripts/cdk-floor/synth.mjs
laazyj: Two honest answers:

**Scaling per-package coverage.** This composed synth is intentionally a *representative* smoke (modelled on the `order-processor` example to avoid hand-writing builder configs that could go stale). It exercises 8/16 packages directly — core, sqs, lambda, sns, cloudwatch — plus cloudformation, iam, logs transitively as peers. The acm/route53/ec2 gap is real.

The exhaustive answer is the per-package floors PR (#156): `cdk-floors:enforce` packs and loads *every* publishable package against its own declared floor (16/16 today, auto-discovers as packages are added). That's the regression guard. So after #156 the division of labour is:
- This smoke (always-on, PR gate): cross-package integration on real CDK; representative.
- `cdk-floors:enforce` (PR gate): exhaustive per-package import check on real CDK at each package's declared floor.

The smoke could grow — add an s3 bucket, an ec2 vpc, an acm cert, a route53 record. Trade-off is more builder configs to maintain. Happy to extend if you'd prefer it touches every builder.

**Policy coverage.** It's `alarmActionsPolicy` because policies-as-Aspects are *the* failure-mode pattern this whole effort exists to prevent — the policy walks the construct tree using `CfnAlarm`/`CfnCompositeAlarm` identification, which is exactly the #146 surface and exactly the class of API #151's lint rule guards against. `alarmNamePolicy` is the same shape (Aspect over the alarm subtree) so it's redundant for this purpose; other policies (s3 lifecycle, etc.) would each add a *different* invariant worth probing. Not "copied from elsewhere" — leaned on as the canonical failure-mode exemplar. Happy to add a second one if there's a specific policy you'd like exercised end-to-end.

scripts/cdk-floor/synth.mjs
laazyj: The alarms assertion is the cross-cutting **Aspect** check. `alarmActionsPolicy` walks the whole construct tree at synth time and decorates every `CfnAlarm`/`CfnCompositeAlarm` — that's exactly the surface that broke in #146, and exactly the kind of API the lint rule in #151 exists to ban. The resource counts (Topic / Queue / Function) prove the system synthesised; the "every alarm wired" check proves the Aspect plumbing did its identification + traversal end-to-end on the pinned CDK.

So alarms aren't universally special — they're the *canonical failure-mode pattern* this effort exists to prevent. If we shipped any other Aspect (e.g. a tag-application policy that traverses every resource), it would deserve the same assertion.

.github/workflows/cdk-floor-suites.yml
laazyj: Per-run, via the `workflow_dispatch` input on line 11 (`cdk-version`), which feeds the `env.CDK_FLOOR` on line 26, which the script reads as `FLOOR`. So a user dispatches the workflow from the Actions tab, fills the version, the script runs against that. The `default: 2.230.0` is a starting point; you'd dispatch with whichever version you wanted to probe (e.g. 2.100.0 to find out "what would break if we lowered everything to here").

Locally the same env var: `CDK_FLOOR=2.100.0 npm run test:cdk-floor-suites --force`.

It's deliberately *not* a set-once value — the script's job is to probe arbitrary versions for floor-research, not enforce a fixed claim. Enforcement of fixed per-package floors lives in #156's `cdk-floors:enforce`.

scripts/cdk-floor-suites.mjs
laazyj: It doesn't honour per-package floors — and that's the central tension this PR shows.

This script pins **one** aws-cdk-lib version workspace-wide and runs all suites. That's the right tool for asking "if we lowered everything to X, what would break?" (floor-research), but it's the *wrong* tool for enforcement once packages have different declared floors.

The per-package enforcement is **`cdk-floors:enforce` (in #156)**, which:
1. Reads `cdk-floors.json` and groups packages by their declared floor.
2. Packs the @composurecdk graph.
3. For each floor group, installs the tarballs against `aws-cdk-lib@<that floor>`.
4. Asserts every package in the group loads.

That's the "don't breach the declared floor" guard, per-package.

So after #156 the three pieces settle into:
- `cdk-floor-test.mjs` (this PR, always-on PR gate): integration smoke on a representative compose() system.
- `cdk-floors:enforce` (#156, PR gate): exhaustive per-package floor enforcement.
- `cdk-floor-suites.mjs` (this PR, manual dispatch): uniform-floor research probe.

That third tool's role narrows substantially once #156 lands. Reasonable to drop `cdk-floor-suites` from this PR and rely on `cdk-floors:enforce` (per-package, correct) + re-running `cdk-floors:establish` for floor-lowering research. See the top-level review reply — happy to do that on this branch if you'd prefer the tighter scope.

Issue #149, opened 2026-05-25
Make `id` parameter optional on createAlarms and defaultAlarmName


laazyj:
### Summary

Both `createAlarms(scope, id, definitions)` and `defaultAlarmName(scope, id, key)` require an `id` parameter that gets baked into:

- The construct id pattern: `${id}${Capitalize(key)}Alarm`
- The default alarm name: `${stackName}/${kebab(id)}/${kebab(key)}`

This locks consumers into the library's specific naming/ID convention. Real-world consumers often have an existing alarm naming convention they need to preserve (to avoid replacing every alarm on first deploy), or are migrating from another framework that produces a flatter structure.

### Proposal

Make the `id` parameter optional on both functions. When omitted:

```ts
// createAlarms(scope, definitions)
//   → construct id: `${Capitalize(key)}` (no id prefix, no Alarm suffix)
//
// defaultAlarmName(scope, key)
//   → `${stackName}/${kebab(key)}`
```

When supplied: existing 3-arg behavior is preserved. This is purely additive — no existing caller breaks.

### Why I'm asking

We're using `@composurecdk/cloudwatch@0.8.0`. The original code had a vendored copy of these utilities that intentionally dropped the `id` parameter (commit message: "Preserve existing CloudFormation construct IDs to avoid alarm replacement on deploy"). That vendored copy is the last thing keeping us from consuming the library directly for `createAlarms` / `defaultAlarmName`.

We considered an `AlarmNameStrategy` policy/strategy interface, but:

- Construct ids are fixed at construction time. CDK Aspects fire too late — by then `node.id` and `node.path` are baked into cross-stack refs and dependency edges. So construct-id naming can't be moved to a synth-time policy. It has to be a parameter on `createAlarms`.
- Alarm names are already covered by `alarmNamePolicy` (synth-time rewriting works because `cfn.alarmName` is just a CFN property). Consumers who want a custom name format can already use that.

So the cross-cutting half (alarm names) has a policy. The construction-time half (construct ids) just needs an optional parameter — no new abstraction.

### Tradeoff

This adds two valid call signatures per function. Slight API-surface growth, but the alternative is forking, which is what we did, and that's worse for everyone. Documenting "if you want hierarchical naming, pass `id`; if you want flat, omit it" is a one-line README addition.

### Related

- #144 (widen `AlarmDefinition.metric` to `IMetric`) — independent, but together the two changes let us delete our vendored fork entirely.
[...]

PR #185, opened 2026-05-31
feat(cloudwatch): per-alarm constructId override

laazyj:
## Summary

Adds an optional, per-definition `constructId` override to alarm creation, so consumers can control the CDK construct id (and therefore the CloudFormation logical ID) of each alarm `createAlarms` produces.

- New optional `constructId?: string` on `AlarmDefinition`.
- New `.constructId(id)` fluent setter on `AlarmDefinitionBuilder`, mirroring the existing `.alarmName(...)`.
- `createAlarms` uses a supplied `constructId` **verbatim** instead of the derived `${id}${Capitalize(key)}Alarm`; an empty string throws a clear error.

When omitted, behaviour is unchanged.

## Why

Resolves the construct-id half of #149. Consumers grandfathering existing stacks (or migrating from another framework) need to preserve their existing CloudFormation logical IDs — a changed construct id changes the logical ID, which **replaces** (drops + recreates) the alarm on first deploy. Construct ids are fixed at construction time, so unlike alarm *names* (already covered by the per-definition `alarmName` field and the scope-wide `alarmNamePolicy`) they cannot be rewritten by a synth-time policy and must be set here.

This deliberately mirrors the existing `alarmName` override rather than introducing a new abstraction. A broader, library-wide construct-id strategy across all builders was considered and **deferred** — there isn't enough signal yet to justify a contract-freezing change, and `CfnResource.overrideLogicalId()` remains the preferred escape hatch in the meantime. See the decision write-up on #149.

### Scope / non-goals

- Direct `createAlarms` path only. Threading overrides through the ~14 service builders is **not** included (deferred).
- No change to `defaultAlarmName` (alarm *names* are already solved two ways).
- No ADR — reuses the existing `alarmName` field pattern.

## Design choices

- **Verbatim** — the supplied id is used exactly as given (no sanitisation), so it can match a legacy logical ID 1:1.
- **Throw on empty** — a present-but-empty `constructId` throws rather than silently falling back.

## Tests

- `constructId` used verbatim; the default formula is not applied; `alarmName` still derives its default independently.
- Empty `constructId` throws.
- Builder propagates `.constructId(...)` into the resolved definition (and leaves it undefined by default).

`npx nx test cloudwatch` (build + typecheck + 91 tests), `npm run lint`, and `npm run format:check` all pass.

Refs #149

### DIFF
--- a/packages/cloudwatch/src/create-alarms.ts
+++ b/packages/cloudwatch/src/create-alarms.ts
@@ -15,11 +15,15 @@ function capitalize(s: string): string {
  * otherwise it is derived from the scope, `id`, and `def.key` via
  * {@link defaultAlarmName}.
  *
+ * Each alarm's construct id is `${id}${Capitalize(key)}Alarm`, unless the
+ * definition supplies an explicit `constructId`, which is used verbatim (e.g.
+ * to preserve an existing logical ID when grandfathering).
+ *
  * @param scope - CDK construct scope.
  * @param id - Base identifier; each alarm's construct id is `${id}${Capitalize(key)}Alarm`.
  * @param definitions - Fully-resolved alarm definitions.
  * @returns A record mapping each definition's key to its created Alarm.
- * @throws If duplicate keys are found in the definitions.
+ * @throws If duplicate keys are found, or a supplied `constructId` is empty.
  */
 export function createAlarms(
   scope: IConstruct,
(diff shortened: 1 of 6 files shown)


packages/cloudwatch/src/create-alarms.ts
```diff
@@ -35,7 +39,11 @@ export function createAlarms(
           `Disable the recommended alarm first, or use a different key.`,
       );
     }
-    alarms[def.key] = def.metric.createAlarm(scope, `${id}${capitalize(def.key)}Alarm`, {
+    if (def.constructId === "") {
+      throw new Error(`Alarm "${def.key}": constructId, when set, must be non-empty.`);
+    }
+    const constructId = def.constructId ?? `${id}${capitalize(def.key)}Alarm`;
```
laazyj, line 45: Shouldn't this be using the constructId sanitiser to protect against invalid values in `def.key` and/or max-length over-runs of the constructed Id?

laazyj (commented):
One open question to clarify

packages/cloudwatch/src/create-alarms.ts
laazyj: Reviewed. Decision: Leave as-is.

- This is not a regression, the existing behaviour exposes errors should `key` include characters invalid for a constructId.
- Length is solved by CDK, it truncates and hashes overly long constructIds silently.
- An existing mechanism (`sanitizeConstructId`) _is_ available in `core` but `cloudwatch` currently doesn't depend on `core` - this is a new dependency.
- There is an existing backlog item to generalise the string sanitisation across the project: #166 . Best to address this gap when the mechanism has stabilised.

PR #188, opened 2026-06-06
feat(constraints): AWS-property constraint catalogue

laazyj:
Implements #166: a synth-time validation catalogue for AWS property strings, so malformed values fail in `cdk synth` at the authoring call site instead of at CloudFormation deploy time (the trigger was an em-dash in a SecurityGroup `GroupDescription`).

Design and the decisions behind it are recorded in **ADR-0010** and the [plan amendments on #166](https://github.com/laazyj/composureCDK/issues/166#issuecomment-4637925493).

## Consumer usage

Most consumers never call a validator directly — the builder enforces it at `build()` / synth:

```ts
import { createSecurityGroupBuilder } from "@composurecdk/ec2";

createSecurityGroupBuilder()
  .vpc(vpc)
  .description("Bastion — SSH entry point") // em-dash
  .build(stack, "Bastion");
// ❌ throws at synth: EC2 SecurityGroup GroupDescription "Bastion — SSH entry point" is invalid.
//    Allowed: ASCII letters, digits, spaces and ._-:/()#,@[]+=&;{}!$*. See <aws-doc>.
```

When you do want to validate a value yourself (tests, manual composition), each package exposes a uniform `constraints` namespace — and you import **only** the package you already use, never the whole catalogue:

```ts
import { constraints } from "@composurecdk/ec2";

constraints.validate.securityGroupDescription("Bastion - SSH"); // ok
constraints.validate.securityGroupName("sg-foo");               // throws: reserved "sg-" prefix
// constraints.sanitize.* is available where a derived value should be rewritten rather than rejected
```

The shape (`constraints.validate.*` / `constraints.sanitize.*`) is identical in every builder package, so discovery is "look at the package that owns the resource." The browsable index of all of them is the generated [`docs/constraints.md`](docs/constraints.md).

Authoring a new constraint (in the owning package) uses the shared mechanism:

```ts
import { stringConstraint, validateString, charSets } from "@composurecdk/cloudformation";

const BUCKET_NAME = stringConstraint({
  name: "S3 Bucket Name",
  charClass: `${charSets.ALNUM}.\\-`,
  minLength: 3,
  maxLength: 63,
  allowed: "lowercase letters, digits, dots and hyphens",
  source: "https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html",
});
export const validateBucketName = (raw: string): void => validateString(raw, BUCKET_NAME);
```

## Public API (kept deliberately small)

- `@composurecdk/cloudformation` — the mechanism: `StringConstraint`, `stringConstraint()`, `validateString`, `sanitizeString`, `charSets` (grouped fragments), `ConstraintNamespace`. Tags stay behind `validateTag` (their `StringConstraint`s are module-private).
- Each builder package — only its `constraints` namespace. The underlying `StringConstraint` definitions and `validate*` functions are module-private.

## Discoverability / keep-it-alive tooling
- **`docs/constraints.md`** — generated by `scripts/gen-constraint-catalogue.mjs` from each package's `constraints` namespace; rows reference the ergonomic `constraints.validate.*` / `constraints.sanitize.*` calls. A `--check` mode is wired into `npm run verify`, so a new constraint can't merge without its catalogue row.
- **`composurecdk/constraint-metadata-required`** ESLint rule — requires every `stringConstraint()` call to set a non-empty `name`, `allowed`, and `source`, catching the gap the type checker can't see (present-but-empty metadata) so every synth-time error stays useful.

## Deliberately out of scope
- Migrating `construct-id` / `alarmName` / `email` onto the mechanism (deferred in ADR-0010).
- The holistic synth-time Aspect (Proposal C) and branded-type migrations remain opt-in/later.

Closes #166

---

Verified locally: full `npm run verify` (`format:check`, `build`, `catalogue:check`, `check:exports`, `lint`, `cdk-floors:check`, and the test suite across 19 projects) passes. Latest `main` is merged in; the ADR was renumbered 0009 → 0010 to avoid a number collision with #187's ADR.


docs/adr/0010-aws-property-constraints.md
```diff
@@ -0,0 +1,55 @@
+# ADR 0010: AWS-property constraints — a catalogue with central mechanism and local data
+
+- **Status:** Accepted
+- **Date:** 2026-06-06
+
+## Context
+
+AWS rejects malformed property strings — character sets beyond an allowed set, over-length values — at deploy time, hours after `cdk synth`. While integration-testing the SecurityGroup builder, an em-dash (`—`) in a `GroupDescription` reached CloudFormation and failed at CREATE*FAILED with *"Character sets beyond ASCII are not supported."\_ A synth-time check would have caught it at the authoring call site.
```
laazyj: "hours after" is subjective - but it is some time after a successful `cdk synth`. Don't make up values, just be accurate.

packages/cloudformation/src/constraints/string-constraint.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * The shared mechanism behind the AWS-property constraint catalogue.
+ *
+ * AWS rejects malformed property strings (bad character sets, over-length
+ * values) at deploy time, hours after `cdk synth`. A {@link StringConstraint}
```
laazyj: As above - the "hours after" is subjective

packages/cloudformation/src/constraints/string-constraint.ts
```diff
@@ -0,0 +1,118 @@
...
+ * captures one AWS property's character-set and length rules as data, so a
+ * builder can fail at synth — at the authoring call site — instead.
+ *
+ * The catalogue is deliberately split: this mechanism lives here, while the
+ * per-resource constraint *data* lives in the package that owns the builder
+ * (e.g. `SECURITY_GROUP_DESCRIPTION` in `@composurecdk/ec2`). Cross-cutting
+ * constraints that apply to every resource — tags — are the exception and
```
laazyj: Don't assume `tags` will be the _only_ cross-cutting resource. Documentation will go stale.

packages/cloudformation/src/index.ts
```diff
@@ -7,5 +7,14 @@ export { singleStack, groupedStacks } from "./strategies.js";
 export { outputs, type OutputDefinition, type OutputDefinitions } from "./outputs.js";
 export { taggedBuilder, type ITaggedBuilder, TAG_OVERRIDE_WARNING_NAME } from "./tagged-builder.js";
 export { applyBuilderTags } from "./apply-builder-tags.js";
-export { validateTag } from "./tag-validator.js";
+export { validateTag, TAG_KEY, TAG_VALUE } from "./tag-validator.js";
```
laazyj: If `TAG_KEY` and `TAG_VALUE` need to be exported then we need to tidy up their naming. Export them in a namespace or provide more descriptive names. These names are ambiguous without their context.

packages/cloudformation/src/index.ts
```diff
@@ -7,5 +7,14 @@ export { singleStack, groupedStacks } from "./strategies.js";
...
 export { tags, type TagDefinitions } from "./tags.js";
+export {
+  type StringConstraint,
+  stringConstraint,
+  validateString,
+  sanitizeString,
+  ALNUM,
+  AWS_NAME_PUNCT,
```
laazyj: There's a lot of new exports for this package. I wonder if there is value in exporting some of them under a namespace? i.e. `char-sets.ALNUM`.

Also, if the constants do need to be exported as part of the package's public API - they need to have better human readable names as changing them later will be difficult.

packages/ec2/src/index.ts
```diff
@@ -37,3 +43,23 @@ export {
   type SecurityGroupBuilderResult,
 } from "./security-group-builder.js";
 export { SECURITY_GROUP_DEFAULTS } from "./security-group-defaults.js";
+export {
+  SECURITY_GROUP_DESCRIPTION,
+  SECURITY_GROUP_NAME,
+  validateSecurityGroupDescription,
+  validateSecurityGroupName,
```
laazyj: Do we really need to export the raw values for these functions _alongside_ their ergonomic exports on L59?

packages/ec2/src/index.ts
```diff
@@ -37,3 +43,23 @@ export {
   type SecurityGroupBuilderResult,
 } from "./security-group-builder.js";
 export { SECURITY_GROUP_DEFAULTS } from "./security-group-defaults.js";
+export {
+  SECURITY_GROUP_DESCRIPTION,
+  SECURITY_GROUP_NAME,
```
laazyj: Why export these values, they can be kept encapsulated within the module and exposed just through the `constraints` namespace.

packages/ec2/src/security-group-constraints.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const SECURITY_GROUP_DESCRIPTION: StringConstraint = stringConstraint({
+  name: "EC2 SecurityGroup GroupDescription",
+  charClass: SG_CHAR_CLASS,
+  maxLength: 255,
+  allowed: SG_ALLOWED,
+  source: SG_SOURCE,
+});
+
+export const SECURITY_GROUP_NAME: StringConstraint = stringConstraint({
```
laazyj: These can remain module private and only accessible via the exported validation functions

docs/constraints.md
```diff
@@ -0,0 +1,20 @@
...
+fails in `cdk synth` at the authoring call site instead of at CloudFormation deploy time.
+The mechanism lives in `@composurecdk/cloudformation`; per-resource entries live in the
+package that owns the builder. See [ADR-0010](adr/0010-aws-property-constraints.md).
+
+Call a constraint's validator from the package that owns it — e.g.
+`import { validateSecurityGroupDescription } from "@composurecdk/ec2"`, or
+`pkg.constraints.validate.*` for the uniform per-package namespace. Importing a service
+package pulls in only that package, never the whole catalogue.
+
+| Constraint                         | Package                        | Constant                     | Length | Allowed                                                                       | AWS reference                                                                              |
```
laazyj: Perhaps more useful than the Constant (which tells the user nothing and I am recommending we do not export), it might be better to reference the ergonomic, namespace-prefixed, function name (or names if both a sanitise and validate are provided for the same constraint). This is genuinely usable and reinforces the ergonomic namespace we inserted to drive discoverability.

packages/cloudformation/src/tag-validator.ts
```diff
@@ -1,63 +1,74 @@
...
+  source: TAG_SOURCE,
+  flags: "u",
+});
+
+// Exported as public catalogue entries — symmetric with per-resource
+// constraints like SECURITY_GROUP_DESCRIPTION. They let callers validate tag
+// strings directly (`validateString(key, TAG_KEY)`) and make tags appear in
+// the generated catalogue (ADR-0010). The `validate*` entry point stays
+// `validateTag`, which layers the empty-key and reserved-prefix rules on top.
+export { TAG_KEY, TAG_VALUE };
```
laazyj: keep these module private

laazyj (commented):
Mostly ok but I think we can tidy up the public API some more. I'm also curious to understand the _consumer_ view of the namespace. Can you include some example consumer code snippets in the PR description?

docs/adr/0010-aws-property-constraints.md
laazyj: Fixed — reworded to "at deploy time, after a successful `cdk synth`" (no invented duration).

---

packages/cloudformation/src/constraints/string-constraint.ts
laazyj: Fixed — same reword here: "at deploy time, after a successful `cdk synth`".

---

packages/cloudformation/src/constraints/string-constraint.ts
laazyj: Reworded to "Cross-cutting constraints that apply to every resource (such as tags) live alongside this mechanism" — no longer implies tags are the only one.

---

packages/cloudformation/src/index.ts
laazyj: Made `TAG_KEY` / `TAG_VALUE` module-private — they're no longer exported. They're reached only through `validateTag`, so the ambiguous names never hit the public surface.

---

packages/cloudformation/src/index.ts
laazyj: Done — `ALNUM` / `AWS_NAME_PUNCT` are now grouped under a single `charSets` export (`charSets.ALNUM`), and `TAG_KEY` / `TAG_VALUE` are no longer exported. The package's new public additions are now just `charSets` plus the mechanism (`StringConstraint`, `stringConstraint`, `validateString`, `sanitizeString`, `ConstraintNamespace`).

---

packages/ec2/src/index.ts
laazyj: Removed — the constants and the `validate*` functions are no longer re-exported. The `constraints` namespace is the only public surface now.

---

packages/ec2/src/index.ts
laazyj: Agreed — encapsulated. `ec2/index.ts` now exports only `constraints`; the definitions and `validate*` functions stay module-private (the builder imports them internally, and the test imports from the module file directly).

---

packages/ec2/src/security-group-constraints.ts
laazyj: Done — `SECURITY_GROUP_DESCRIPTION` / `SECURITY_GROUP_NAME` are now module-private `const`s, reached only via the exported `validate*` functions (which the `constraints` namespace surfaces).

---

docs/constraints.md
laazyj: Reworked the catalogue. The "Constant" column is gone; it now lists the ergonomic, namespace-prefixed references per constraint, with separate Validate / Sanitize columns (e.g. `constraints.validate.securityGroupDescription`). Since constants are no longer exported, the generator discovers entries by reading each package's `constraints` namespace.

---

packages/cloudformation/src/tag-validator.ts
laazyj: Done — kept module-private (no longer exported from the package index).

---

PR #189, opened 2026-06-06
feat(neptune): add Amazon Neptune cluster builder

laazyj:
Closes #141.

Adds `@composurecdk/neptune` — a fluent builder for Amazon Neptune clusters wrapping the `@aws-cdk/aws-neptune-alpha` L2 with well-architected defaults, composition support, and recommended alarms.

## Design choices

The issue sketched a **two-builder split** (cluster + instance, RDS-shaped). Building against the real alpha L2 changed the calculus and this PR ships a **single cluster builder**:

- **Single cluster builder, not a cluster/instance split.** The L2 `DatabaseCluster` *owns* its instances (it requires an `instanceType` and creates the writer/readers from the `instances` count), and does not expose them as separate constructs. In-region read replicas are added via the instance count (`.instances(n)`); cross-region scale-out is a [Neptune Global Database](https://docs.aws.amazon.com/neptune/latest/userguide/neptune-global-database.html) — separate read-only *clusters* per region (no alpha L2 yet), not extra instances. A separate instance builder would only uniquely serve *heterogeneous* in-region readers, which AWS guidance doesn't push (failover targets should match the writer), so it is **deferred** until that need is concrete.
- **`instanceType` is required, not defaulted.** Defaulting it would create surprise cost; the builder throws if it is unset.
- **Audit logging is coherent end-to-end.** Only `LogType.AUDIT` is CloudWatch-exportable. Audit *export* only emits data once audit logging is *enabled* in the cluster parameter group, whose `family` must match the engine version. So the builder auto-creates an audit-log-enabled cluster parameter group with the family derived from the engine version — parallel to how `createVpcBuilder` auto-creates a flow-log group. Override via `.clusterParameters({...})` or supply your own `.clusterParameterGroup()` (mutually exclusive).
- **`allowAccessFrom(peer)` for access (answers issue Q3).** Because IAM auth is a default, a principal needs *both* network ingress and an IAM `connect` grant. `allowAccessFrom` applies `connections.allowDefaultPortFrom(peer)` and `grantConnect(peer)` in one declaration and accepts a `Ref`, so access is data in the `compose()` graph. When IAM auth is disabled it opens only the network path (the IAM grant would be inert).
- **Recommended alarms** (CPU, request-queue backlog, buffer-cache hit ratio, replica lag, and a serverless-only capacity alarm) run through the shared `@composurecdk/cloudwatch` helpers. Neptune is absent from the CloudWatch recommended-alarms table, so thresholds cite the Neptune metrics guidance and the Well-Architected Neptune lens.
- **Alpha dependency as a peer.** `@aws-cdk/aws-neptune-alpha` is a `peerDependency` (consumers pin it); its version-locked relationship to `aws-cdk-lib` is recorded as the package's `cdk-floors.json` floor.

## What's included

- `packages/neptune/` — builder, secure defaults, parameter-group + alarm modules, README, and a unit test suite.
- Registered in `@composurecdk/module-compat` (dual ESM/CJS resolution) and `cdk-floors.json`.
- `ComposureCDK-NeptuneGraphStack` example (serverless cluster + bastion demonstrating `allowAccessFrom`) with a synth test and a liveness smoke test. Teardown for the CI cycle is handled by `cleanDeskPolicy` so the example keeps its production-shaped stateful defaults.

## Verification

`npm run verify` passes (build, `check:exports`, lint, format, cdk-floors, and the full test suite across 20 projects). The changed code was reviewed with the `simplify` skill.


docs/adr/0010-neptune-cluster-builder.md
```diff
@@ -0,0 +1,81 @@
...
+## Decision
+
+Ship Neptune as a **single `createClusterBuilder`** in `@composurecdk/neptune`.
+
+- **One builder, both topologies.** Serverless (`InstanceType.SERVERLESS` +
+  `serverlessScalingConfiguration`) and provisioned (`InstanceType.R6G_LARGE`,
+  `.instances(n)`) are the same builder. `instanceType` is required — the
+  builder throws if it is unset rather than defaulting it and creating surprise
+  cost. A separate instance builder is deferred until there is a concrete need
+  for heterogeneous replicas (issue #141 Q2).
```
laazyj: I would like to see if we can address the "heterogeneous replicas" now rather than later. 
1. Research how a Neptune cluster should be scaled out in AWS documentation (especially Well Architected Framework). If the pattern implies declaring additional read-replicas (i.e. in other regions) separately from specifying the number of instances in the cluster, then this might confirm `InstanceBuilder` as a real need.
2. If we do identify this as a real need, then we should look to have the `ClusterBuilder` using the `InstanceBuilder` internally to supply the simpler (# instances) API too.

docs/adr/0010-neptune-cluster-builder.md
```diff
@@ -0,0 +1,81 @@
...
+  `cloudwatchLogsExports: [AUDIT]` default coherent end-to-end. `.clusterParameters({...})`
+  merges extra parameters; a user-supplied group disables the auto group (the
+  two are mutually exclusive).
+- **`allowAccessFrom(peer)` for access.** Because IAM authentication is a
+  default, a principal needs both network ingress and an IAM `connect` grant.
+  `allowAccessFrom` applies `connections.allowDefaultPortFrom(peer)` and
+  `grantConnect(peer)` in one declaration and accepts a `Resolvable`, so the
+  grant is data inside `compose()` rather than `afterBuild` glue. This answers
+  issue #141 Q3 and adopts the combined-grant shape the real-world commenter
+  validated.
```
laazyj: I don't think this change warrants an ADR. It is confined to the implementation of a single resource class in a single package. It has no cross-package, architectural, impact.

It's possible - in future - that the `allowAccessFrom()` decision might become an architectural lynchpin but that is yet to be proven.

packages/examples/src/neptune-graph-app.ts
```diff
@@ -0,0 +1,97 @@
...
+ * VPC-only and needs no internet egress. The bastion therefore has no
+ * outbound route either, so SSM Session Manager will not connect as-is — for
+ * an interactively reachable jump-box, switch the VPC to NAT/egress subnets
+ * (or add interface VPC endpoints for SSM) and place the bastion there. The
+ * bastion exists here to demonstrate the access-grant wiring, which the
+ * post-deploy smoke test verifies on the live security groups.
+ *
+ * `deletionProtection` and `removalPolicy` are overridden to allow the CI
+ * deploy/destroy cycle to tear the cluster down — production stacks keep the
+ * `RETAIN` defaults.
```
laazyj: The overrides of deletion protection & removal policy should be implemented in `cleanDeskPolicy` keeping the example as an exemplar of a real, albeit small, system.

packages/examples/src/neptune-graph-app.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+export function createNeptuneGraphApp(app = new App()) {
+  const stack = new Stack(app, "ComposureCDK-NeptuneGraphStack");
+
+  const { alerts } = compose(
+    {
+      alerts: createTopicBuilder().displayName("Neptune Alerts"),
```
laazyj: We don't need to demonstrate the alert SNS notifications again. There are plenty of examples. Shrink the example footprint by omitting it.

packages/examples/test/smoke/neptune-graph.smoke.mjs
```diff
@@ -0,0 +1,110 @@
...
+    } else {
+      fail(`${STACK} — IAM database authentication is OFF (security default violated)`);
+    }
+
+    const auditExported = (cluster.EnabledCloudwatchLogsExports ?? []).includes("audit");
+    if (auditExported) {
+      pass(`${STACK} — audit logs are exported to CloudWatch`);
+    } else {
+      fail(`${STACK} — audit log export is OFF (logging default violated)`);
+    }
```
laazyj, line 186: Much of the above tests seem to replicate the unit tests of the builder itself + the snapshot of the example app. This is unnecessary overhead to update the smoke test for small changes to configuration. Unless these properties cannot be verified via the snapshot let's aim to reduce the smoke test foot print to verifying the running system, not all of its configuration.

Continuing that thought, verifying the configuration of the bastion SG is _not_ a robust smoke test. Consider switching to the following, real and functional, path: SSM to bastion (or tunnel via), query the Neptune server port, run a basic Cypher or Gremlin status query to verify the cluster is running,

packages/neptune/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+The defaults are exported as `CLUSTER_DEFAULTS` for visibility and testing.
+
+### Audit-log parameter group
+
+Audit log _export_ only emits data once audit logging is _enabled_ in the cluster parameter group. So the builder auto-creates a cluster parameter group with `neptune_enable_audit_log = "1"` (parallel to how `createVpcBuilder` auto-creates a flow-log group), with the family derived from the configured engine version. Add or override parameters with `.clusterParameters({...})`, or supply your own group with `.clusterParameterGroup(myGroup)` (mutually exclusive with `.clusterParameters()`).
+
+## Granting access — `allowAccessFrom`
+
+Because IAM authentication is on by default, a principal needs both a network path (security-group ingress) and an IAM `connect` grant to reach the cluster. `allowAccessFrom(peer)` does both in one declaration, and accepts a `Ref` so the grant lives inside `compose()` rather than in post-build glue:
```
laazyj, line 85: What if the user disables IAM? Does `allowAccessFrom` adapt for that? Is it compatible for that use case?

laazyj (commented):
Good first draft. Some open questions and suggestions to be considered

packages/examples/test/smoke/neptune-graph.smoke.mjs
```diff
@@ -0,0 +1,45 @@
...
+    const clusterId = clusterResources[0].PhysicalResourceId;
+
+    const { DBClusters: clusters } = aws(
+      "neptune",
+      "describe-db-clusters",
+      "--db-cluster-identifier",
+      clusterId,
+      "--output",
+      "json",
+    );
```
laazyj, line 113: This is still insufficient for a smoke test. It just shows that a cluster is defined not that it is alive and accessible within the example application.

I recommend we look at enabling SSM access, via the bastion, to query the graphdb itself using OpenCypher to check its health. This extends the usefulness of the example application to show how to make a usable Neptune DB as well. If it requires defining security groups then we can also demonstrate the ec2 SecurityGroupBuilder in the same app.

laazyj (commented):
Smoke tests are still too shallow

packages/examples/test/smoke/neptune-graph.smoke.mjs
laazyj: Done in [`6a7d1fa`](https://github.com/laazyj/composureCDK/commit/6a7d1fa) — the smoke test now proves the graph is reachable and queryable, not just defined.

- **Example:** the bastion gets a closed-egress SG via `createSecurityGroupBuilder`, and the VPC gains `ssm`/`ssmmessages`/`ec2messages` interface endpoints so Session Manager reaches it without a NAT gateway (keeps the VPC isolated). Cross-component wiring adds exactly two egress rules to the bastion SG — `:8182` to Neptune (via `allowAccessFrom`) and `:443` to the endpoints — so least privilege stays visible.
- **Smoke test:** `ssm send-command` runs a script on the bastion that issues a SigV4-signed OpenCypher `RETURN 1 AS health` query against the IAM-authenticated cluster endpoint, then polls the invocation. Signing is pure Python stdlib (reads the instance-role creds from IMDSv2) rather than `awscurl`/`boto3`, since the isolated bastion has no internet egress to install anything.

One judgement call worth flagging: I went with **VPC interface endpoints** rather than a NAT gateway so the example stays isolated and cost-free, which is also why the query is signed in-process instead of via `awscurl`. If you'd prefer the NAT + `awscurl` shape for readability, happy to switch. This path is deploy-only, so it's exercised by the `deploy-test` workflow rather than CI unit tests.

---

packages/examples/test/smoke/neptune-graph.smoke.mjs
```diff
@@ -0,0 +1,183 @@
...
+req = urllib.request.Request(
+    "https://" + host + ":" + port + "/openCypher", data=body.encode(), method="POST",
+    headers={"Content-Type": "application/x-www-form-urlencoded",
+             "X-Amz-Date": amzdate, "X-Amz-Security-Token": st,
+             "Authorization": authorization})
+resp = urllib.request.urlopen(req, timeout=15)
+out = resp.read().decode()
+print("HTTP", resp.status, out)
+sys.exit(0 if (resp.status == 200 and '"health"' in out) else 1)
+`;
```
laazyj, line 94: Why not just install `awscurl` using `pip`?

packages/examples/src/neptune-graph-app.ts
```diff
@@ -0,0 +1,121 @@
...
+  // SSM Session Manager reachability without a NAT gateway: interface
+  // endpoints for the three services the agent needs. Each opens :443 to the
+  // bastion only, which also adds the matching egress rule on the bastion's
+  // closed-egress SG.
+  addSsmEndpoints(result.network.vpc, result.bastion.instance);
+
+  return { stack };
+}
+
+function addSsmEndpoints(vpc: IVpc, bastion: Instance): void {
```
laazyj, line 111: There's something funky here. Why does this need to be declared outside of the system composition? What API surfaces are missing from composurecdk libraries that forces this approach?

laazyj (commented):
Needs more work. The smoke test has recreated awscurl inline when it could reuse. The networking rules to allow SSM access need to be folded into the system compose function call. If API surface gaps exist then raise a blocking GitHub issue describing the use case.

PR #198, opened 2026-06-09
feat(core): expose resolved builder configuration on results (resolvedProps)

laazyj:
Closes the mechanism portion of #122 (and unblocks #118/#123/#124).

## Problem

Builders need to inspect a property that configured a **sibling** component, but the CDK L2 construct doesn't re-expose it as a public readonly member — CDK treats props as a write-only struct. The motivating case: the SQS↔Lambda visibility-timeout invariant (a queue's `visibilityTimeout` should be ≥ 6× the consumer's `timeout`). `Queue` doesn't expose `visibilityTimeout`; only `QueueProps` carries it, so a consumer receiving the queue via `ref()` can only emit a contextual reminder, never a real comparison (see the "not enforced" note on `sqsEventSource`).

Three facts make the value unreachable today: the construct doesn't re-expose it; the builder getter reads `props` *before* defaults are merged and lives on the builder (which siblings never receive); and the invariant spans two components, so only the consumer — which already knows its own side — can check it.

## Approach (per the [analysis on #122](https://github.com/laazyj/composureCDK/issues/122#issuecomment-4663832379), recommended Option A)

A builder result exposes the resolved configuration it handed to its CDK construct under a uniform `resolvedProps` field. The merged object already exists inside every `build()` (`{ ...DEFAULTS, ...userProps }`, refs resolved) — it was simply being discarded. This extends the existing "build results must be complete" tenet from constructs to resolved configuration.

```ts
const orders = createQueueBuilder().visibilityTimeout(Duration.seconds(30));
const result = orders.build(stack, "Orders");
result.resolvedProps.visibilityTimeout; // Duration | undefined — the Queue construct never exposes this
```

A consumer holding a `ref()` to the result can now run a real, token-guarded check instead of a reminder.

## Changes

- **core** — `ResolvedProps<P>` (and `Resolved<T>`): `Readonly<{ [K in keyof P]?: Resolved<P[K]> }>` — every key optional and read-only, with any `Resolvable<T>` collapsed to its resolved `T`. Exported from the package root.
- **sqs** — `QueueBuilderResult.resolvedProps: ResolvedProps<QueueProps>` (the #118 unblocker).
- **lambda** — `FunctionBuilderResult.resolvedProps: ResolvedProps<FunctionProps>`, which exercises the `Resolvable` collapse via the widened `role` prop.
- **docs** — [ADR-0011](docs/adr/0011-resolved-props-on-results.md) records the result-type contract and the alternatives weighed (curated named fields, capability channel, decorator, status quo); `architecture.md` and the ADR index updated.

## Design notes

- **Semantics: post-default, post-resolve, *not* token-resolved.** A value may still be an unresolved CDK `Token`; consumers guard with `Token.isUnresolved`, exactly as `QueueBuilder`'s existing `warnIfLowMaxReceiveCount` does.
- **Typed off the CDK props, not the builder props** — builder-specific keys (`recommendedAlarms`) are destructured out before the merge, and widened keys collapse back to the construct's type.
- **Originates in `build()`** (returned as a shallow copy), deliberately not a decorator: a decorator can't see `mergedProps`, which is internal to `build()`. ADR-0011 §Alternatives covers this.
- **Scope:** full merged prop set, not a curated subset — `resolvedProps` is the floor; a builder may *additionally* promote a specific value to a named field.

## Scope / follow-up

This lands the mechanism plus the two motivating builders. Rolling `resolvedProps` out to the remaining ~33 `*BuilderResult` types is a mechanical follow-up (mirrors how ADR-0010 shipped a central mechanism plus first adopters). The actual `sqsEventSource` invariant check that consumes this is #123/#124.

## Verification

`npm run verify` green (build, `check:exports`, lint, test across all 19 projects). New tests: `resolvedProps` defaults/override/write-only-prop/by-reference on `QueueBuilder`; defaults + `ref()`-role collapse on `FunctionBuilder`; type-level optional/collapse/readonly assertions on `ResolvedProps` in core.

> [!NOTE]
> One maintainer call flagged in the #122 analysis: `resolvedProps` exposes the *full* prop bag, including values with no cross-component use and potentially sensitive ones (e.g. a Lambda's `environment`). It's in-process only and never serialised, but it is a public contract widening. If that's unwelcome, the fallback is curated named fields under a single namespace — happy to adjust.

https://claude.ai/code/session_013qoPQbmyWj5sGHpNWTT3jQ

---
_Generated by [Claude Code](https://claude.ai/code/session_013qoPQbmyWj5sGHpNWTT3jQ)_

### DIFF
--- a/docs/adr/0011-resolved-props-on-results.md
+++ b/docs/adr/0011-resolved-props-on-results.md
@@ -1,0 +1,128 @@
+# ADR 0011: Expose a builder's resolved configuration on its result
+
+- **Status:** Accepted
+- **Date:** 2026-06-09
+
+## Context
+
+Builders frequently need to inspect a property that configured a _sibling_
+component, but the CDK L2 construct does not expose it as a public readonly
+member. The value lives only on the `*Props` interface, which CDK treats as a
+write-only struct ([CDK design guidelines][cdk-guidelines]) — once the
+construct is built, the value is unrecoverable from the construct alone.
+
+The case that surfaced this (laazyj/composureCDK#122, blocking #118): the
+SQS↔Lambda visibility-timeout invariant — a source queue's `visibilityTimeout`
+should be ≥ 6× the consumer function's `timeout`. The `Queue` construct does
+not expose `visibilityTimeout`; only `QueueProps` carries it. A consumer that
+receives the queue via `ref()` therefore cannot perform a real comparison — it
+can only emit a contextual reminder (see the "not enforced" note on
+`sqsEventSource`). This is a general gap: _any_ cross-component invariant that
+depends on a write-only prop of a dependency is un-checkable today.
+
+Three facts about the codebase frame the decision:
+
+1. **The merged configuration already exists.** Every builder's `build()`
+   computes `mergedProps = { ...DEFAULTS, ...userProps }` (with any
+   `Resolvable` resolved against `context`) and hands it to the construct. It
+   is then discarded.
+2. **Results already carry more than the primary construct** — `FunctionBuilderResult.eventSources`,
+   `RuleBuilderResult.targets`, `RoleBuilderResult.inlinePolicies`,
+   `TopicBuilderResult.subscriptions`. But every "extra" today is a construct
+   or a resolved CDK object; **no result exposes the merged scalar config**
+   (a `Duration`, a number, an enum) that the construct withholds.
+3. **The builder getter is not a substitute.** `queue.visibilityTimeout()`
+   reads `props` _before_ defaults are merged and lives on the builder, which
+   a sibling never receives — siblings receive the result via `context`.
+
+[architecture.md](../architecture.md) already states the tenet "**if a builder
+creates it, the result should expose it**." Resolved configuration is part of
+that completeness, not just constructs.
+
+## Decision
+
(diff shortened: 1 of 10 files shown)


laazyj:
## Rejecting this approach — closing

After weighing the graph-integrity analysis above against the value delivered, we're not taking `resolvedProps` in any form:

- **As a general mechanism it's too large an architectural change for limited value, and it weakens a core guarantee** — that the `compose(...)` dependency map is the single source of truth for the system's edges. Exposing the full prop bag (incl. construct handles) lets a consumer wire CloudFormation edges the graph never declared.
- **Narrowing it to scalar-only props doesn't redeem it** — it makes the mechanism *less* idiomatic (a bespoke `ScalarProps` projection of CDK props that exists solely to dodge the handle-leak), while still adding a result-contract surface to every builder.

The motivating need (the SQS↔Lambda visibility-timeout check, #123) does **not** actually require this. The premise behind #122 — "the value is unrecoverable from the construct" — holds only for the **L2**. The **L1 `CfnQueue.visibilityTimeout` is a public getter**, readable at/after synth via the standard escape hatch, with no cross-package dependency and no graph edge created (it's a scalar read, not a handle wire). #123 will be progressed on that basis instead.

Closing this PR and #122. #195 will be reworked onto the L1/synth-time approach.

PR #201, opened 2026-06-10
feat(ec2): VPC interface-endpoint builder (#194)

laazyj:
## What this is

Implements \`createInterfaceEndpointBuilder\` — a first-class \`compose()\` component for VPC interface endpoints (AWS PrivateLink). Closes the API surface proposed in [#194](https://github.com/laazyj/composureCDK/issues/194).

## Design

One builder maps **1:1 to a CDK \`InterfaceVpcEndpoint\`** (one \`service\` per endpoint). Security group has two mutually-exclusive modes:

- **BYO** — \`.securityGroups([...])\` with SGs you fully manage (sibling \`SecurityGroupBuilder\`s): full ingress/egress/port control. No SG created; \`securityGroup\` absent from the result.
- **Managed shortcut** — omit \`.securityGroups()\`; the builder auto-creates a closed SG, exposes it on the result, and \`.allowDefaultPortFrom(peer)\` opens ingress on the service's default port (\`service.port\`, 443 for AWS services).

Combining BYO \`.securityGroups()\` with \`.allowDefaultPortFrom()\` throws. \`open\` is always \`false\` — explicit ingress only. Group several endpoints under one access policy by pointing them at the same SG.

```
createInterfaceEndpointBuilder()
  .vpc(Resolvable<IVpc>)                              // required
  .service(InterfaceVpcEndpointAwsService | custom)   // required (fluent prop)
  .subnets(...) / .privateDnsEnabled(...)             // fluent props off InterfaceVpcEndpointProps
  .securityGroups(Resolvable<ISecurityGroup>[])        // BYO  ─┐ mutually
  .allowDefaultPortFrom(Resolvable<IPeer>, desc?)      // managed ┘ exclusive
  .recommendedAlarms({...} | false)                   // alarm config
  .addAlarm("key", alarm => ...)                      // custom alarms
// result: { endpoint, securityGroup?, alarms }
```

## Defaults

| Property | Default | Rationale |
| --- | --- | --- |
| `privateDnsEnabled` | `true` | Resolves service hostnames to endpoint ENIs without application changes (Well-Architected SEC 6). |
| `open` | `false` | Always hard-coded — SG ingress is always explicit. |

## Recommended alarms

Sourced from the [AWS CloudWatch recommended-alarms reference](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Best_Practice_Recommended_Alarms_AWS_Services.html#PrivateLinkEndpoints):

| Alarm | Metric | Default threshold |
| --- | --- | --- |
| `packetsDropped` | `AWS/PrivateLinkEndpoints PacketsDropped` (Sum, 1 min) | > 0 over 5 consecutive 1-min windows |

Configurable via \`.recommendedAlarms({...})\` / \`.addAlarm()\`. Threshold default of 0 signals any sustained drop; workloads sending packets > 8,500 bytes (PrivateLink MTU) should raise it.

### 1. Minimalist single service (managed shortcut)

```ts
ssm: createInterfaceEndpointBuilder()
  .vpc(ref<VpcBuilderResult>("network").get("vpc"))
  .service(InterfaceVpcEndpointAwsService.SSM)
  .subnets({ subnetType: SubnetType.PRIVATE_ISOLATED })
  .allowDefaultPortFrom(ref<SecurityGroupBuilderResult>("bastionSg").get("securityGroup")),
// result.ssm = { endpoint, securityGroup, alarms }
```

### 2. SSM access — multiple endpoints, one shared SG

One endpoint per builder; share a single \`SecurityGroupBuilder\`. The three SSM services (\`SSM\`, \`SSM_MESSAGES\`, \`EC2_MESSAGES\`) are declared explicitly — no \`SSM_ACCESS_SERVICES\` helper.

### 3. Complex / custom service (BYO SG, non-443, ingress + egress)

```ts
partner: createInterfaceEndpointBuilder()
  .vpc(ref<VpcBuilderResult>("network").get("vpc"))
  .service(new InterfaceVpcEndpointService("com.amazonaws.vpce.eu-west-1.vpce-svc-0abc123", 8443))
  .securityGroups([ref<SecurityGroupBuilderResult>("endpointSg").get("securityGroup")]),
// result.partner = { endpoint, alarms }  (no managed securityGroup in BYO mode)
```

## Changes

- \`packages/ec2/src/interface-endpoint-builder.ts\` — builder + alarm wiring
- \`packages/ec2/src/interface-endpoint-defaults.ts\` — Well-Architected defaults
- \`packages/ec2/src/interface-endpoint-alarm-config.ts\` — alarm config type
- \`packages/ec2/src/interface-endpoint-alarm-defaults.ts\` — alarm defaults const
- \`packages/ec2/src/interface-endpoint-alarms.ts\` — alarm creation
- \`packages/ec2/src/index.ts\` — exports (removed \`SSM_ACCESS_SERVICES\`)
- \`packages/ec2/README.md\` — full builder documentation

### DIFF
--- a/packages/ec2/src/interface-endpoint-alarms.ts
+++ b/packages/ec2/src/interface-endpoint-alarms.ts
@@ -1,0 +1,83 @@
+import { Duration } from "aws-cdk-lib";
+import { type Alarm, ComparisonOperator, Metric, Stats } from "aws-cdk-lib/aws-cloudwatch";
+import { type InterfaceVpcEndpoint } from "aws-cdk-lib/aws-ec2";
+import type { IConstruct } from "constructs";
+import type { AlarmDefinition } from "@composurecdk/cloudwatch";
+import { AlarmDefinitionBuilder, createAlarms, resolveAlarmConfig } from "@composurecdk/cloudwatch";
+import type { InterfaceEndpointAlarmConfig } from "./interface-endpoint-alarm-config.js";
+import { INTERFACE_ENDPOINT_ALARM_DEFAULTS } from "./interface-endpoint-alarm-defaults.js";
+
+const PACKETS_DROPPED_PERIOD = Duration.minutes(1);
+const PACKETS_DROPPED_PERIOD_LABEL = `${String(PACKETS_DROPPED_PERIOD.toMinutes())} minute`;
+
+function endpointMetric(
+  endpoint: InterfaceVpcEndpoint,
+  metricName: string,
+  statistic: string,
+  period: Duration,
+): Metric {
+  return new Metric({
+    namespace: "AWS/PrivateLinkEndpoints",
+    metricName,
+    dimensionsMap: { "VPC Endpoint Id": endpoint.vpcEndpointId },
+    statistic,
+    period,
+  });
+}
+
+export function resolveEndpointAlarmDefinitions(
+  endpoint: InterfaceVpcEndpoint,
+  config: InterfaceEndpointAlarmConfig | undefined,
+): AlarmDefinition[] {
+  if (config?.enabled === false) return [];
+
+  const definitions: AlarmDefinition[] = [];
+
+  if (config?.packetsDropped !== false) {
+    const cfg = resolveAlarmConfig(
+      config?.packetsDropped,
+      INTERFACE_ENDPOINT_ALARM_DEFAULTS.packetsDropped,
+    );
+    definitions.push({
+      key: "packetsDropped",
+      alarmName: cfg.alarmName,
+      metric: endpointMetric(endpoint, "PacketsDropped", Stats.SUM, PACKETS_DROPPED_PERIOD),
+      threshold: cfg.threshold,
+      comparisonOperator: ComparisonOperator.GREATER_THAN_THRESHOLD,
+      evaluationPeriods: cfg.evaluationPeriods,
+      datapointsToAlarm: cfg.datapointsToAlarm,
+      treatMissingData: cfg.treatMissingData,
+      description:
+        `VPC interface endpoint is dropping packets — possible endpoint service unhealthy, ` +
+        `security group blocking traffic, or packets exceeding the 8,500-byte PrivateLink MTU. ` +
+        `Threshold: > ${String(cfg.threshold)} (sum) over ` +
+        `${String(cfg.evaluationPeriods)} x ${PACKETS_DROPPED_PERIOD_LABEL}.`,
+    });
+  }
+
+  return definitions;
+}
+
+/**
+ * Creates AWS-recommended CloudWatch alarms for a VPC interface endpoint,
+ * merging recommended definitions with any custom alarm builders.
+ *
+ * @see https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Best_Practice_Recommended_Alarms_AWS_Services.html#PrivateLinkEndpoints
+ */
+export function createInterfaceEndpointAlarms(
+  scope: IConstruct,
+  id: string,
+  endpoint: InterfaceVpcEndpoint,
+  config: InterfaceEndpointAlarmConfig | false | undefined,
+  customAlarms: AlarmDefinitionBuilder<InterfaceVpcEndpoint>[] = [],
+): Record<string, Alarm> {
+  if (config === false) return {};
+
+  const enabled = config?.enabled ?? INTERFACE_ENDPOINT_ALARM_DEFAULTS.enabled;
+  if (!enabled) return {};
+
+  const recommended = resolveEndpointAlarmDefinitions(endpoint, config);
+  const custom = customAlarms.map((b) => b.resolve(endpoint));
+
+  return createAlarms(scope, id, [...recommended, ...custom]);
+}
--- a/packages/ec2/src/interface-endpoint-alarms.ts
+++ b/packages/ec2/src/interface-endpoint-alarms.ts
@@ -25,7 +25,7 @@ function endpointMetric(
   });
 }
 
-export function resolveEndpointAlarmDefinitions(
+function resolveEndpointAlarmDefinitions(
   endpoint: InterfaceVpcEndpoint,
   config: InterfaceEndpointAlarmConfig | undefined,
 ): AlarmDefinition[] {
--- a/packages/ec2/src/interface-endpoint-builder.ts
+++ b/packages/ec2/src/interface-endpoint-builder.ts
@@ -1,0 +1,259 @@
+import { type Alarm } from "aws-cdk-lib/aws-cloudwatch";
+import {
+  InterfaceVpcEndpoint,
+  type InterfaceVpcEndpointProps,
+  type IPeer,
+  type ISecurityGroup,
+  type IVpc,
+  Port,
+  type SecurityGroup,
+} from "aws-cdk-lib/aws-ec2";
+import { type IConstruct } from "constructs";
+import { COPY_STATE, type Lifecycle, resolve, type Resolvable } from "@composurecdk/core";
+import { type ITaggedBuilder, taggedBuilder } from "@composurecdk/cloudformation";
+import { AlarmDefinitionBuilder } from "@composurecdk/cloudwatch";
+import { createSecurityGroupBuilder } from "./security-group-builder.js";
+import { INTERFACE_ENDPOINT_DEFAULTS } from "./interface-endpoint-defaults.js";
+import type { InterfaceEndpointAlarmConfig } from "./interface-endpoint-alarm-config.js";
+import { createInterfaceEndpointAlarms } from "./interface-endpoint-alarms.js";
+
+/**
+ * Configuration properties for the interface-endpoint builder.
+ *
+ * Lifts three CDK props off the props object:
+ * - `vpc` — supplied via {@link IInterfaceEndpointBuilder.vpc | .vpc()} so it
+ *   can accept a {@link Resolvable} for cross-component wiring.
+ * - `securityGroups` — supplied via
+ *   {@link IInterfaceEndpointBuilder.securityGroups | .securityGroups()} so
+ *   each can be a {@link Resolvable} (typically a sibling
+ *   `SecurityGroupBuilder`).
+ * - `open` — always `false`; ingress is explicit (see the builder docs).
+ */
+export interface InterfaceEndpointBuilderProps extends Omit<
+  InterfaceVpcEndpointProps,
+  "vpc" | "securityGroups" | "open"
+> {
+  /**
+   * Configuration for AWS-recommended CloudWatch alarms.
+   *
+   * By default, the builder creates recommended alarms with sensible
+   * thresholds. Individual alarms can be customized or disabled. Set to
+   * `false` to disable all alarms.
+   *
+   * @see https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Best_Practice_Recommended_Alarms_AWS_Services.html#PrivateLinkEndpoints
+   */
+  recommendedAlarms?: InterfaceEndpointAlarmConfig | false;
+}
+
+/**
+ * The build output of an {@link IInterfaceEndpointBuilder}.
+ *
+ * `securityGroup` is present only in **managed mode** — i.e. when the caller
+ * did *not* supply `.securityGroups(...)`, so the builder auto-created one. It
+ * is exposed so a peer's *egress* rule can `ref` it. In **BYO mode** it is
+ * `undefined`: the caller already holds refs to the security groups they
+ * passed in.
+ */
+export interface InterfaceEndpointBuilderResult {
+  endpoint: InterfaceVpcEndpoint;
+  securityGroup?: SecurityGroup;
+  alarms: Record<string, Alarm>;
+}
+
+interface AccessSpec {
+  readonly peer: Resolvable<IPeer>;
+  readonly description?: string;
+}
+
+/**
+ * A fluent builder for a single VPC interface endpoint (AWS PrivateLink).
+ *
+ * Unlike raw CDK — where interface endpoints exist only as a post-build
+ * `vpc.addInterfaceEndpoint(...)` call whose security group is never exposed —
+ * this builder is a first-class {@link compose} component. It maps 1:1 to a
+ * CDK `InterfaceVpcEndpoint` (one `service` per endpoint); group several into
+ * one access policy by pointing them at the same security group.
+ *
+ * **Security group, two modes:**
+ * - *BYO* — call {@link IInterfaceEndpointBuilder.securityGroups | .securityGroups([...])}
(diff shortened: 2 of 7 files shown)


packages/ec2/src/interface-endpoint-builder.ts
```diff
@@ -0,0 +1,197 @@
...
+ * Lifts three CDK props off the props object:
+ * - `vpc` — supplied via {@link IInterfaceEndpointBuilder.vpc | .vpc()} so it
+ *   can accept a {@link Resolvable} for cross-component wiring.
+ * - `securityGroups` / `open` — the builder *owns* the endpoint's security
+ *   group so it can expose it on the result and let peers `ref` it. Ingress is
+ *   declared with {@link IInterfaceEndpointBuilder.allowDefaultPortFrom}.
+ */
+export type InterfaceEndpointBuilderProps = Omit<
+  InterfaceVpcEndpointProps,
+  "vpc" | "securityGroups" | "open"
```
laazyj, line 34: What are we losing by not allowing BYO `securityGroups` for the consumer? 

Does the raw CDK interface allow users to more precisely manage the SGs with ingress (and presumably) egress access to the endpoint, while our builder only allows defining `Peer`s with _ingress_ rules?

I'm not against a more constrained, opinionated, API that optimises for the common use case - but I want to understand the trade-offs we're making.

packages/ec2/src/interface-endpoint-builder.ts
```diff
@@ -0,0 +1,197 @@
...
+      this.#configureSg,
+    );
+    for (const rule of this.#access) {
+      securityGroup.addIngressRule(resolve(rule.peer, context), Port.tcp(443), rule.description);
+    }
+
+    const endpoint = new InterfaceVpcEndpoint(scope, id, {
+      ...INTERFACE_ENDPOINT_DEFAULTS,
+      ...this.props,
+      service: this.props.service as IInterfaceVpcEndpointService,
```
laazyj: Why do we need to coerce service to `IInterfaceVpcEndpointService`?

packages/ec2/src/interface-endpoint-builder.ts
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  /**
+   * Opens ingress on the endpoint's owned security group from `peer` on the
+   * endpoint's default port (443). Mirrors CDK's
+   * `endpoint.connections.allowDefaultPortFrom(...)` and the SG builder's
+   * `addIngressRule`. The peer can be a concrete {@link IPeer} or a
+   * {@link Ref} to a sibling component's security group.
+   */
+  allowDefaultPortFrom(peer: Resolvable<IPeer>, description?: string): this {
```
laazyj, line 148: If a user wants to add access to a port other than `443` they use `configureSecurityGroup`? Demonstrate in a reply to this comment.

packages/ec2/src/interface-endpoint-builder.ts
```diff
@@ -0,0 +1,197 @@
...
+      securityGroup.addIngressRule(resolve(rule.peer, context), Port.tcp(443), rule.description);
+    }
+
+    const endpoint = new InterfaceVpcEndpoint(scope, id, {
+      ...INTERFACE_ENDPOINT_DEFAULTS,
+      ...this.props,
+      service: this.props.service as IInterfaceVpcEndpointService,
+      vpc: resolvedVpc,
+      securityGroups: [securityGroup],
+      open: false,
```
laazyj, line 233: I understand this is a secure default. Demonstrate in a reply to this comment how a consumer would get similar functionality to `open: true` using this builder.

Secondly - shouldn't this value be derived from `INTERFACE_ENDPOINT_DEFAULTS`, or is this hard-coded so that it is _not_ overridable?

packages/ec2/src/interface-endpoints-builder.ts
```diff
@@ -0,0 +1,178 @@
...
+  InterfaceVpcEndpointAwsService.SSM_MESSAGES,
+  InterfaceVpcEndpointAwsService.EC2_MESSAGES,
+];
+
+/**
+ * Configuration shared across every endpoint in the bundle. Per-service
+ * settings (the service identity itself) are added via `.services()` /
+ * `.ssmAccess()`; the access edge via `.allowDefaultPortFrom()`.
+ */
+export interface InterfaceEndpointsBuilderProps {
```
laazyj: It is an odd choice to have two builders for essentially the same thing, but one with a plural name. 

It looks like there are multiple motivations for this pattern:
- Simplify the compose map arguments where multiple services are required to achieve a single functional endpoint (SSM)
- Solve for the shared SG across multiple endpoints when the singular builder doesn't allow BYO SG.

It feels janky to have the two options. What would it look like to fold them into one?

And is the ability to define _all_ of your endpoints in a single component - even if unrelated - an anti pattern that we are encouraging?

```
compose({
  vpc: createVpcBuilder(),
  everything: createInterfaceEndpointsBuilder()
    .vpc(ref("vpc", v -> v.vpc)
    .services([https, ssm, ssh, ftp, http, smtp, etc]),
},
);
```

laazyj (commented):
Consider further use cases than just SSM -> bastion and how the API design works (or doesn't) for those.

packages/ec2/src/interface-endpoint-builder.ts
laazyj: We don't — it's cast-noise and should go.

`this.props` is `Partial<…>`, so `this.props.service` is typed `IInterfaceVpcEndpointService | undefined`, and `...this.props` carries that optionality into the literal where `service` is required. The `undefined` guard a few lines up already proves it's defined; capturing it in a narrowed local removes the cast — and the explicit `service:` key is then fully redundant, since `...this.props` already spreads it:

```ts
const { service } = this.props;
if (service === undefined) {
  throw new Error(`InterfaceEndpointBuilder "${id}" requires a service. …`);
}
// …
new InterfaceVpcEndpoint(scope, id, {
  ...INTERFACE_ENDPOINT_DEFAULTS,
  ...this.props,      // already carries `service`
  vpc: resolvedVpc,
  securityGroups: [securityGroup],
  open: false,
});
```

Will fix.

---

packages/ec2/src/interface-endpoint-builder.ts
laazyj: Honest trade-offs of the owned-SG model as drafted:

Raw CDK `securityGroups` lets you attach SGs you **fully own** — ingress *and* egress, any port, any peer, and the same SG shared across several endpoints. My `allowDefaultPortFrom` surface only expresses **ingress on :443 from an `IPeer`**. So as drafted we lose: egress rules on the endpoint ENI, non-443 ports (custom PrivateLink services aren't always 443), and a shared/BYO SG that lives as its own `ref`-able component. What it buys is the one-liner for the SSM-style common case.

The thing is — we **already** have that precise SG surface: `SecurityGroupBuilder` has full `addIngressRule`/`addEgressRule`/`addSelfIngress` + `ref` support. Re-inventing a constrained rule API on the endpoint builder duplicates it and is the source of every limitation above.

So I'd rather **add `.securityGroups(Resolvable<ISecurityGroup>[])` (BYO)** and only auto-create + expose an SG when none is supplied, keeping `.allowDefaultPortFrom(peer)` as sugar over the auto-created one:

```ts
// Power case — full control via the SG builder you already have:
endpointSg: createSecurityGroupBuilder().vpc(ref<VpcBuilderResult>("net").get("vpc"))
  .description("ECR endpoint").addIngressRule(ref("tasksSg").get("securityGroup"), Port.tcp(443)),
ecr: createInterfaceEndpointBuilder().vpc(ref(...))
  .service(InterfaceVpcEndpointAwsService.ECR_DOCKER)
  .securityGroups([ref<SecurityGroupBuilderResult>("endpointSg").get("securityGroup")]),

// Common case — unchanged one-liner, builder owns + exposes the SG:
ssm: createInterfaceEndpointBuilder().vpc(ref(...)).service(...).allowDefaultPortFrom(ref("bastionSg")...),
```

That keeps the opinionated common case terse while restoring full ingress/egress/port/peer control for everything else — and it's the more architecturally consistent answer (composition is data; the SG is a first-class node). This is the same change that resolves the non-443 (#discussion below), `open: true`, and fold-the-two-builders threads — they're one revised design.

---

packages/ec2/src/interface-endpoint-builder.ts
laazyj: As drafted, yes — via `configureSecurityGroup`, which is clunky and the weakest part of this API.

Under the BYO-SG change I proposed on the props thread, a custom-port PrivateLink service is just the `SecurityGroupBuilder` you already pass in — no special hatch:

```ts
endpointSg: createSecurityGroupBuilder()
  .vpc(ref<VpcBuilderResult>("net").get("vpc"))
  .description("custom PrivateLink svc")
  .addIngressRule(ref<SecurityGroupBuilderResult>("client").get("securityGroup"), Port.tcp(8443)),

endpoint: createInterfaceEndpointBuilder()
  .vpc(ref<VpcBuilderResult>("net").get("vpc"))
  .service(new InterfaceVpcEndpointService("com.amazonaws.vpce.…", 8443))
  .securityGroups([ref<SecurityGroupBuilderResult>("endpointSg").get("securityGroup")]),
```

`.allowDefaultPortFrom(peer)` stays as :443 sugar for the AWS-service common case; anything off-443 uses the SG builder. That lets me **delete `configureSecurityGroup`** entirely — it only existed to work around the owned-SG-with-no-BYO limitation.

---

packages/ec2/src/interface-endpoint-builder.ts
laazyj: CDK's `open: true` means "add an ingress rule on :443 from the **whole VPC CIDR**" to the endpoint SG. The equivalent here is one `allowDefaultPortFrom` against the VPC CIDR (or a BYO SG with that rule):

```ts
.allowDefaultPortFrom(
  Peer.ipv4(ref<VpcBuilderResult>("net").map((r) => r.vpc.vpcCidrBlock)),
  "Open to VPC (open:true equivalent)",
)
```

On the second question — good catch, and it's a genuine bug: `open` is in `INTERFACE_ENDPOINT_DEFAULTS` **and** hard-coded `false` in the literal *after* the spread, so the default is dead (shadowed). It is deliberately not overridable: because the builder manages the SG, letting `open: true` through would silently bolt a VPC-wide :443 rule onto *our* SG behind the user's back. So the fix is to **remove `open` from `INTERFACE_ENDPOINT_DEFAULTS`** (it's an implementation detail of SG ownership, not a tunable) and keep it hard-coded `false` with a comment. `privateDnsEnabled` stays in the defaults object since it's genuinely user-tunable. The current state — half in defaults, half hard-coded — is the inconsistency you're pointing at.

---

packages/ec2/src/interface-endpoints-builder.ts
laazyj: Agreed, the singular/plural split is janky. Fold into one builder where the singular is just sugar:

```ts
createInterfaceEndpointBuilder()
  .vpc(ref<VpcBuilderResult>("net").get("vpc"))
  .service(svc)        // sugar for .services([svc])
  .services([a, b, c]) // additive
  .allowDefaultPortFrom(ref("clientSg").get("securityGroup")) // or .securityGroups([...]) BYO
// result: { endpoints: Record<string, InterfaceVpcEndpoint>, securityGroup?: SecurityGroup }
//         (securityGroup present only when auto-created)
```

`SSM_ACCESS_SERVICES` stays as an exported preset list you spread into `.services(...)`; no second builder.

On the anti-pattern worry — that's the right question, and the answer is the cohesion boundary: **one builder = one shared SG = one access policy.** `.services([...])` means "these endpoints share *this* SG and *this* ingress." So your `everything: [https, ssh, ftp, smtp, …]` example is actively discouraged *because* those want different access policies — they belong in separate components with separate SGs. Grouping is by shared access, not "dump every endpoint in one node." The legitimate multi-service-one-policy cases are exactly the ones worth bundling:

- **SSM/Session Manager** — `ssm` + `ssmmessages` + `ec2messages`, identical :443-from-bastion.
- **ECR image pulls** — `ecr.api` + `ecr.dkr` (+ the S3 gateway endpoint), shared task-SG ingress.
- **Lambda-in-VPC** reaching `secretsmanager` + `kms`, shared function-SG ingress.

(Re your review summary about looking past SSM→bastion: those three are the use cases I stress-tested this against, plus the custom-port / whole-VPC / egress-controlled cases on the other threads — all of which is why I'm leaning to BYO-SG + a single folded builder rather than the constrained ingress-only surface in this draft.)

---

packages/ec2/src/interface-endpoint-builder.ts
```diff
@@ -0,0 +1,231 @@
...
+
+/**
+ * The interface endpoints required for AWS Systems Manager / Session Manager
+ * reachability in a NAT-free, isolated VPC. Exported as plain data — one
+ * endpoint component is created per entry, all sharing a single security
+ * group (see the SSM example in the README).
+ *
+ * @see https://docs.aws.amazon.com/systems-manager/latest/userguide/setup-create-vpc.html
+ */
+export const SSM_ACCESS_SERVICES: readonly IInterfaceVpcEndpointService[] = [
```
laazyj: This is a convenience for a single use case and doesn't need to be part of our API. Consumers requiring SSM access will be able to determine the 3 services they require.

Please remove `SSM_ACCESS_SERVICES`

packages/ec2/src/interface-endpoint-defaults.ts
```diff
@@ -0,0 +1,24 @@
+import type { InterfaceVpcEndpointProps } from "aws-cdk-lib/aws-ec2";
+
+/**
+ * Defaults applied to every interface endpoint built with
+ * {@link createInterfaceEndpointBuilder}.
+ *
+ * DRAFT (#194): only the load-bearing default is present — the full
+ * Well-Architected JSDoc citations land with the non-draft PR.
```
laazyj: Needs to be corrected before final version

laazyj (commented):
I'm happy with the shape. Let's productions this:
- Remove SSM_ACCESS_SERVICES convenience const
- Research and apply Well Architected Framework recommendations
- Research and apply AWS recommended alarms
- Sweep documentation and comments for "draft" comments
- Review README content for publication readiness

packages/ec2/src/interface-endpoint-alarms.ts
```diff
@@ -0,0 +1,83 @@
...
+  return new Metric({
+    namespace: "AWS/PrivateLinkEndpoints",
+    metricName,
+    dimensionsMap: { "VPC Endpoint Id": endpoint.vpcEndpointId },
+    statistic,
+    period,
+  });
+}
+
+export function resolveEndpointAlarmDefinitions(
```
laazyj: This can be module-private

laazyj (commented):
1 minor change

PR #205, opened 2026-06-11
fix(lambda): guard token-valued alarm thresholds (#196)

laazyj:
## Summary

`FunctionBuilder.build()` threw when `timeout` was a token-valued `Duration` (e.g. derived from `CfnParameter.valueAsNumber`): the recommended duration alarm called `toMilliseconds()` **unconditionally, ahead of the `duration !== false` escape hatch**, and that conversion throws on a cross-unit token. Closes #196.

Rather than patch the one call site, this generalises the underlying mechanism and sweeps every alarm builder in the library for the same shape.

## What changed

**`feat(cloudwatch)` — new shared guard.** `resolveAlarmThresholdBasis` resolves the numeric basis for any alarm whose threshold is *derived* from a possibly-token input. When the input is an unresolved token it annotates the scope with a standardized, acknowledgeable warning and returns `undefined` (signalling the caller to omit the alarm) instead of throwing or rendering a garbage threshold. The token predicate defaults to `Token.isUnresolved` and is overridable for wrapper types like `Duration` that carry their own check. It lives in `@composurecdk/cloudwatch` because every alarm builder already depends on it.

**`fix(lambda)` — adopt the guard + fix ordering.** Both derived-threshold alarms now:
1. evaluate their enabling boolean (`config?.x !== false`) **before** any per-alarm computation, so `recommendedAlarms({ duration: false })` is a reliable escape hatch even when the conversion would throw; and
2. route the `timeout` / `reservedConcurrentExecutions` derivation through the shared guard. A token-valued input omits the alarm and warns (the concurrency alarm previously wouldn't throw but would silently `Math.round(token * pct)` into a meaningless threshold — also fixed).

## Holistic coverage

I audited **every** alarm-producing file across all packages for both failure modes — (1) escape-hatch misordering, (2) threshold arithmetic/unit-conversion on a token-capable builder prop:

| Package | Result |
|---|---|
| lambda | **Fixed** (this PR) |
| acm, apigateway, budgets, cloudfront, ec2, events, route53, s3, sns, sqs | **Clean** — all thresholds are verbatim config pass-throughs (no arithmetic on props); enabling booleans correctly precede computation |
| neptune | **Considered, no change** — the `serverlessDatabaseCapacity` default does `maxCapacity * fraction`, but `@aws-cdk/aws-neptune-alpha` validates `maxCapacity`'s range **without** a token guard, so a token `maxCapacity` throws during cluster construction — the arithmetic is unreachable through the supported API. Adding a guard would be untestable dead code, so it was left alone. |

So lambda was the only genuinely token-exposed site; the shared helper is in place for the next derived-threshold alarm (e.g. the SQS visibility-timeout work in #123).

## Tests

- `cloudwatch`: unit tests for the helper — concrete / converted / undefined / token-number / token-`Duration`, including that the guard short-circuits before `resolve()` runs.
- `lambda`: `build()` succeeds with a token-valued timeout and the duration alarm is omitted + warns; the `duration: false` / `concurrentExecutions: false` escape hatches suppress without warning; same for token-valued `reservedConcurrentExecutions`.

Full `verify` gate (build, `check:exports`, lint, test across all packages) passes.

### DIFF
--- a/packages/lambda/src/function-alarms.ts
+++ b/packages/lambda/src/function-alarms.ts
@@ -9,7 +9,12 @@ import {
 import type { Function as LambdaFunction, FunctionProps } from "aws-cdk-lib/aws-lambda";
 import type { IConstruct } from "constructs";
 import type { AlarmDefinition, AlarmName } from "@composurecdk/cloudwatch";
-import { AlarmDefinitionBuilder, createAlarms, resolveAlarmConfig } from "@composurecdk/cloudwatch";
+import {
+  AlarmDefinitionBuilder,
+  createAlarms,
+  resolveAlarmConfig,
+  resolveAlarmThresholdBasis,
+} from "@composurecdk/cloudwatch";
 import type {
   FunctionAlarmConfig,
   PercentageAlarmConfig,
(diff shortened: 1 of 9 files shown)


packages/lambda/src/function-alarms.ts
```diff
@@ -104,44 +109,74 @@ export function resolveFunctionAlarmDefinitions(
...
-      datapointsToAlarm: cfg.datapointsToAlarm,
-      treatMissingData: cfg.treatMissingData,
-      description: `Lambda function p99 duration is approaching the configured timeout. Threshold: > ${String(threshold)}ms (${String(cfg.thresholdPercent * 100)}% of ${String(timeoutMs)}ms timeout).`,
+  // The duration alarm's threshold is a percentage of the function timeout, so
+  // it can only be derived from a concrete `Duration`. Honour the `duration !==
+  // false` escape hatch *before* touching the timeout, so disabling the alarm
+  // always works. The shared helper then guards the unit conversion: a
+  // token-valued timeout (e.g. threaded through a `CfnParameter`) cannot be
+  // resolved to a millisecond threshold at synth time — `toMilliseconds()`
+  // would throw — so it warns and returns `undefined`, omitting the alarm.
```
laazyj: This (and the next comment on the other alarm) is a lot of text, can it be reduced while keeping the important points

laazyj (commented):
Minor reduction in overlong comments pls

packages/lambda/src/function-alarms.ts
laazyj: Tightened both comments in b20a850 — the duration block is now 5 lines and the concurrency block 3, keeping the escape-hatch ordering and token-guard rationale.

---

PR #221, opened 2026-06-20
feat(dynamodb): add @composurecdk/dynamodb with Table and TableV2 builders

laazyj:
## Summary

Adds the `@composurecdk/dynamodb` package, implementing the revised two-builder design captured in [#121's plan](https://github.com/laazyj/composureCDK/issues/121#issuecomment-4757053415) and the [naming/no-ADR amendment](https://github.com/laazyj/composureCDK/issues/121#issuecomment-4757197851) (supersedes the closed draft #197).

It ships **two thin builders** sharing alarms and well-architected defaults rationale, one per CDK construct:

| Factory | Construct | CFN resource | Use for |
|---|---|---|---|
| `createTableV2Builder()` | `TableV2` | `AWS::DynamoDB::GlobalTable` | New tables (**recommended**) |
| `createTableBuilder()` | `Table` (classic) | `AWS::DynamoDB::Table` | `importSource` (S3 bulk import), parity |

### Why two builders, not one
`TablePropsV2` is not a superset of `TableProps` — most notably the `encryption` prop collides under the same key with an **incompatible type** (`TableEncryption` enum vs `TableEncryptionV2` class), which the Proxy-over-CDK-props builder model (ADR-0001) cannot unify without abandoning the thin pass-through. `createTableBuilder()` deliberately keeps the unqualified name = classic `Table` to match every CDK user's prior; `TableV2` owns the README's "recommended" framing, not the bare factory name.

## What's included

- **Two builders**, both `taggedBuilder`-based and `Lifecycle`-implementing, returning every construct they create (`table`, `tableStreamArn`, `alarms`) via a shared `TableBuilderResultBase`.
- **Well-architected defaults** (`TABLE_DEFAULTS` + `TABLE_V2_DEFAULTS`), each JSDoc-linked to its AWS source: on-demand billing, AWS-managed KMS encryption, point-in-time recovery, deletion protection.
- **AWS-recommended CloudWatch alarms** (system errors, read/write throttling) via a shared `ITable`-typed alarm path — `ITableV2 extends ITable`, so one module serves both.
- **DynamoDB Streams** surfaced as `tableStreamArn` for `ref()` wiring of a future `dynamoEventSource`.
- **Usage docs** in the package README, leading with V2, with a classic section and the no-in-place-migration warning.

## Notable correctness details

- Classic builder keeps the **ADR-0009 sibling yields** (`billingMode`↔capacity, `encryption`↔`encryptionKey`); TableV2 needs none (single helper-object props), so a plain spread suffices.
- TableV2's `tableStreamArn` getter returns a CFN token even with no stream, so the V2 result **guards on the `dynamoStream` prop** to keep `tableStreamArn` `undefined` without a stream — matching the classic builder's contract.
- TableV2 encodes deletion protection + PITR **per-replica** (`Replicas[]`); tests assert against `Replicas[0]`.
- `aws-cdk-lib` floor pinned to **2.178.0**, the release that introduced the L2 `pointInTimeRecoverySpecification` prop both defaults rely on (verified absent in 2.177.0); registered in `cdk-floors.json`.
- `module-compat`: added `@composurecdk/dynamodb` peer + ESM/CJS probes for both `createTableBuilder` and `createTableV2Builder`.

## Testing

- `table-builder.test.ts`, `table-v2-builder.test.ts`: defaults, ADR-0009 yields, streams, copy-state, per-replica V2 encoding.
- `table-alarms.test.ts`: parametrized over **both** builders to prove the shared alarm path.
- Full suite green: lint, format, `cdk-floors:check`, `check:exports`, and tests across all 21 projects (module-compat resolution included).

Closes #121

https://claude.ai/code/session_018jYSQRVE7HSoWH3HfMNbsQ

---
_Generated by [Claude Code](https://claude.ai/code/session_018jYSQRVE7HSoWH3HfMNbsQ)_


packages/dynamodb/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Which builder should I use?
+
+For **new tables, prefer `createTableV2Builder()`**. A single-region `TableV2` bills the same as a classic table, but it future-proofs the one irreversible part of the decision: you can add cross-region replicas later without replacing the table. `ITableV2 extends ITable`, so a built `TableV2` works everywhere an `ITable` is expected (IAM grants, `DynamoEventSource`) — composition wiring is identical for both builders.
+
+Reach for the classic `createTableBuilder()` when you need an `importSource` (S3 bulk import is V1-only, with no `TableV2` equivalent) or parity with existing classic tables.
+
+> ⚠️ **`Table` and `TableV2` are different CloudFormation resources** (`AWS::DynamoDB::Table` vs. `AWS::DynamoDB::GlobalTable`). CloudFormation **cannot migrate between them in place** — swapping the construct on an existing table is a resource replacement, which deletes and recreates the table. Choose deliberately up front.
+
+#### Why two builders instead of one?
```
laazyj: The "Why two builders" section is unnecessary. It is self-evident from the description above.

packages/dynamodb/README.md
```diff
@@ -0,0 +1,201 @@
...
+| Deletion protection    | `deletionProtection: true`                      | `deletionProtection: true`                 | Blocks table deletion (API, console, or `cdk destroy`) until explicitly disabled. Override with `.deletionProtection(false)` for ephemeral tables.                                                                                                                                                |
+
+The defaults are exported for visibility and testing:
+
+```ts
+import { TABLE_DEFAULTS, TABLE_V2_DEFAULTS } from "@composurecdk/dynamodb";
+```
+
+> **TableV2 per-replica encoding:** in `AWS::DynamoDB::GlobalTable`, `BillingMode`, `SSESpecification`, and `StreamSpecification` are top-level, but `DeletionProtectionEnabled` and `PointInTimeRecoverySpecification` are **per-replica** (under `Replicas[]`). The builder sets them once and CDK applies them to each replica; assert against `Replicas[0]` in synth tests.
```
laazyj: This detail about TableV2 replicas is unnecessary. Our behaviour is the expectation.

packages/dynamodb/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Defaults that yield to a sibling you set (classic builder)
```
laazyj: This section (defaults yielding) is unnecessary. The behaviour is both as-expected and sufficiently covered in prop JSDoc.

packages/dynamodb/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+result.tableStreamArn; // string — feed into a Lambda DynamoEventSource, etc.
+result.table; // the table itself, which a DynamoEventSource consumes
+```
+
+Neither construct exposes a distinct stream construct — the stream is an attribute of the table — so the result exposes `tableStreamArn` directly. It is `undefined` when no stream is configured, on both builders, so a `ref()` consumer can branch on its presence.
+
+## Recommended Alarms
+
+Both builders create the same [AWS-recommended CloudWatch alarms](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Best_Practice_Recommended_Alarms_AWS_Services.html#DynamoDB) by default (the alarm path is shared and typed to `ITable`, which both constructs satisfy). No alarm actions are configured — access alarms from the build result to add SNS topics or other actions.
```
laazyj: The detail about the alarm path is unnecessary detail

packages/dynamodb/src/index.ts
```diff
@@ -0,0 +1,16 @@
...
+  type TableBuilderProps,
+  type TableBuilderResult,
+} from "./table-builder.js";
+export {
+  createTableV2Builder,
+  type ITableV2Builder,
+  type TableV2BuilderProps,
+  type TableV2BuilderResult,
+} from "./table-v2-builder.js";
+export { type TableBuilderResultBase } from "./table-result.js";
```
laazyj: Why is `TableBuilderResultBase` exported? This is an internal implementation detail of the package - we should not make long-term guarantees of current inheritance-based implementation.

packages/dynamodb/src/table-alarms.ts
```diff
@@ -0,0 +1,161 @@
...
+    definitions.push({
+      key: "systemErrors",
+      alarmName: cfg.alarmName,
+      // metricSystemErrorsForOperations sums SystemErrors across the given
+      // operations into a single MathExpression — the per-operation dimension
+      // means there is no single plain Metric for "errors on this table".
+      metric: table.metricSystemErrorsForOperations({
+        period: METRIC_PERIOD,
+        operations: SYSTEM_ERROR_OPERATIONS,
+      }) as MathExpression,
```
laazyj, line 77: Is the cast to `MathExpression` necessary?

laazyj (commented):
Minor improvements before merging

packages/dynamodb/src/table-alarms.ts
laazyj: Yes — the cast is necessary. `metricSystemErrorsForOperations()` is declared to return `cloudwatch.IMetric`, but `AlarmDefinition.metric` is typed `AlarmMetric = Metric | MathExpression` (the concrete classes that carry `createAlarm`, per the `AlarmMetric` doc). An `IMetric` isn't assignable to that union, so removing the cast fails to compile. At runtime the helper returns a `MathExpression` (it sums per-operation `SystemErrors` into a metric-math expression), so `as MathExpression` is the accurate narrowing. The inline comment above it notes the same rationale.

---

PR #229, opened 2026-06-21
feat(examples): add DynamoDB table builder example stacks

laazyj:
Add two example stacks demonstrating the @composurecdk/dynamodb builders,
which had no prior example coverage.

- ComposureCDK-EventStoreStack (event-store-app.ts) uses the recommended
  createTableV2Builder: GSI, TTL, and a change stream feeding a Lambda
  projector wired via the raw-CDK DynamoEventSource escape hatch (no
  dynamoEventSource helper exists yet). Tunes a recommended alarm, adds a
  custom alarm, and routes all alarms to an SNS topic via alarmActionsPolicy.
  Global-table replicas are opt-in behind the ddbReplicaRegions context flag
  so the default CI deploy stays single-region and cheap.

- ComposureCDK-LedgerTableStack (ledger-table-app.ts) uses the classic
  createTableBuilder with provisioned capacity and a customer-managed KMS
  key, demonstrating defaults yielding to mutually-exclusive siblings
  (ADR-0009).

Registers both in bin/app.ts, adds the @composurecdk/dynamodb peer dep,
documents them in the README, and adds unit/snapshot tests plus a
post-deploy smoke test for the stream projection path.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_013hfZgjxdTnssoaqNLaegV4


packages/examples/src/event-store-app.ts
```diff
@@ -0,0 +1,152 @@
...
+              }),
+          ),
+        ),
+    },
+    { alerts: [], events: [], projector: ["events"] },
+  ).build(stack, "EventStore");
+
+  alarmActionsPolicy(stack, {
+    defaults: { alarmActions: [new SnsAction(alerts.topic)] },
+  });
```
laazyj: Don't need the alarmActionsPolicy and SnsTopic. This is demonstrated sufficiently elsewhere

packages/examples/test/smoke/event-store.smoke.mjs
```diff
@@ -0,0 +1,79 @@
+import { findStackResources, pollUntil, resolveLambdaLogGroup } from "./_helpers.mjs";
```
laazyj, line 1: Why no smoke test for ledger-table-app?

packages/examples/src/event-store-app.ts
laazyj: Done — removed the SNS topic and `alarmActionsPolicy` from the event-store example (and the `createTopicBuilder`/`SnsAction`/`alarmActionsPolicy` imports). The table's recommended + custom alarms are still created but left unrouted; the JSDoc now points to order-processor-app for the alarm-action wiring. Unit test and snapshot updated (asserts zero SNS topics now). Pushed in dfabb72.

---

packages/examples/test/smoke/event-store.smoke.mjs
laazyj: Added one in dfabb72 — `packages/examples/test/smoke/ledger-table.smoke.mjs`. It `describe-table`s to assert the live table is encrypted with the customer-managed KMS key (the example's headline feature), then round-trips an item (`put-item` + consistent `get-item`) to prove the provisioned, CMK-encrypted table works end to end.

One thing worth flagging: the CI OIDC smoke-test role had **no DynamoDB data-plane permissions**, so this — and the existing event-store stream smoke test (`put-item`) — would have failed in `deploy-test`. I added a `DynamoDBData` statement (`PutItem`/`GetItem`/`DescribeTable`, ARN-scoped to the sandbox account's tables, mirroring the `SQSSendMessage` pattern) in `.github/cloudformation/github-oidc-role.yml`, documented in `docs/ci.md`. That role change has to be deployed to the sandbox account before the next `deploy-test` run.

---

PR #231, opened 2026-06-21
feat(lambda): add dynamoEventSource() and stream IteratorAge alarm

laazyj:
## Summary

Completes the DynamoDB event-source remainder deferred from #118. The SQS half
(`sqsEventSource()` + `FunctionBuilder.addEventSource()`) shipped there;
`dynamoEventSource()` and the stream-only `IteratorAge` alarm were parked until
a DynamoDB builder existed to `ref()` against. #121 (merged as #221) landed
`@composurecdk/dynamodb`, unblocking this. Kinesis (#120) remains.

The source-`kind` seam built for the SQS work made this almost entirely
additive — the compiler forced the new kind into the dispatch maps.

## What's included

- **`dynamoEventSource(table, props?)`** — a `Resolvable`-aware factory mirroring
  `sqsEventSource`: accepts a concrete `ITable` or a `ref()` to a sibling, and
  applies secure defaults via `DEFAULT_DYNAMO_EVENT_SOURCE_PROPS`
  (`startingPosition: LATEST`, `reportBatchItemFailures: true`, `EventCount` ESM
  metrics), each overridable through `props`. At build time the source binds onto
  the function's least-privilege role (`grantStreamRead`).
- **Per-mapping alarms reused** — DynamoDB streams emit the same
  `FailedInvokeEventCount` / `DroppedEventCount` as SQS, so both kinds now share
  one `ESM_EVENT_COUNT_SPECS` array.
- **New `iteratorAge` contextual alarm** — `AWS/Lambda IteratorAge` (Maximum,
  dimensioned on `FunctionName`), created once whenever a stream source is
  attached. Unlike the per-mapping alarms it is function-level, so it sits in its
  own branch alongside `duration` / `concurrentExecutions`, not the per-mapping
  loop. Default `> 60000 ms` for 3 consecutive minutes; tune/disable via the new
  `eventSourceIteratorAge` field.

## Notable details

- **No `aws-cdk-lib` floor change** — `DynamoEventSource` and the `metricsConfig`
  prop are gated by the same `MetricType` feature already at the lambda floor
  (2.168.0); `cdk-floors:check` is unchanged.
- **No new package dependency** — the factory takes `Resolvable<ITable>` from
  `aws-cdk-lib/aws-dynamodb`, not `@composurecdk/dynamodb`, keeping lambda
  decoupled (exactly as `sqsEventSource` depends on `IQueue`, not
  `@composurecdk/sqs`).
- The table must have a stream enabled or CDK's `DynamoEventSource.bind()` throws
  at build time; documented in the factory JSDoc and README.
- `module-compat` unaffected — new exports are additive.

## Testing

- `packages/lambda/test/event-sources.test.ts`: factory defaults (`LATEST`,
  batch-failure, EventCount), `props` override, `ref` deferral, synth of the
  event source mapping + `grantStreamRead`, per-mapping alarms for DynamoDB, and
  the `iteratorAge` alarm (present for a stream source; absent for SQS-only and
  bare escape-hatch sources; one per function; threshold override; disable).
- Full gate green on the new base (`origin/main` incl. the dynamodb lint fix):
  build, `check:exports`, lint, format, `cdk-floors:check`, tests across all 21
  projects.

https://claude.ai/code/session_018jYSQRVE7HSoWH3HfMNbsQ

---
_Generated by [Claude Code](https://claude.ai/code/session_018jYSQRVE7HSoWH3HfMNbsQ)_

### DIFF
--- a/packages/lambda/src/alarm-config.ts
+++ b/packages/lambda/src/alarm-config.ts
@@ -119,4 +119,24 @@ export interface FunctionAlarmConfig {
    * @see https://aws.amazon.com/blogs/compute/introducing-new-event-source-mapping-esm-metrics-for-aws-lambda/
    */
   eventSourceDroppedEvents?: AlarmConfig | false;
+
+  /**
+   * Alarm when the function is falling behind a stream event source.
+   *
+   * Contextual: a single alarm is created when at least one stream event
+   * source (currently DynamoDB streams) is attached via
+   * {@link IFunctionBuilder.addEventSource}. Unlike the per-mapping alarms
+   * above, `IteratorAge` is a function-level metric, so there is one alarm per
+   * function (keyed `iteratorAge`) regardless of how many stream sources are
+   * attached.
+   *
+   * The threshold is an absolute age in milliseconds (not a percentage).
+   *
+   * Metric: `AWS/Lambda IteratorAge`, statistic Maximum, period 1 minute,
+   * dimensioned on `FunctionName`. Default threshold: > 60000 ms (60s) for 3
+   * consecutive minutes.
+   *
+   * @see https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Best_Practice_Recommended_Alarms_AWS_Services.html#Lambda
+   */
+  eventSourceIteratorAge?: AlarmConfig | false;
 }
--- a/packages/lambda/src/alarm-config.ts
+++ b/packages/lambda/src/alarm-config.ts
@@ -125,18 +125,21 @@ export interface FunctionAlarmConfig {
    *
    * Contextual: a single alarm is created when at least one stream event
    * source (currently DynamoDB streams) is attached via
-   * {@link IFunctionBuilder.addEventSource}. Unlike the per-mapping alarms
-   * above, `IteratorAge` is a function-level metric, so there is one alarm per
-   * function (keyed `iteratorAge`) regardless of how many stream sources are
-   * attached.
+   * {@link IFunctionBuilder.addEventSource}. Unlike the per-mapping
+   * {@link FunctionAlarmConfig.eventSourceFailedInvocations} /
+   * {@link FunctionAlarmConfig.eventSourceDroppedEvents} alarms, `IteratorAge`
+   * is a function-level metric, so there is one alarm per function (keyed
+   * `iteratorAge`) regardless of how many stream sources are attached.
    *
    * The threshold is an absolute age in milliseconds (not a percentage).
    *
    * Metric: `AWS/Lambda IteratorAge`, statistic Maximum, period 1 minute,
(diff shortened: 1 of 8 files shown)


packages/lambda/src/alarm-config.ts
```diff
@@ -119,4 +119,24 @@ export interface FunctionAlarmConfig {
...
    */
   eventSourceDroppedEvents?: AlarmConfig | false;
+
+  /**
+   * Alarm when the function is falling behind a stream event source.
+   *
+   * Contextual: a single alarm is created when at least one stream event
+   * source (currently DynamoDB streams) is attached via
+   * {@link IFunctionBuilder.addEventSource}. Unlike the per-mapping alarms
+   * above, `IteratorAge` is a function-level metric, so there is one alarm per
```
laazyj: Don't rely on positional qualifiers ("alarms above") in the JSDoc. This can be displayed - i.e. in IDEs - out of the context of the source file.

packages/lambda/src/alarm-defaults.ts
```diff
@@ -73,4 +74,17 @@ export const FUNCTION_ALARM_DEFAULTS: FunctionAlarmDefaults = {
...
+   * The consumer falling behind the stream risks data loss once records age
+   * past the stream's 24h retention. Alarm on sustained lag — 60s of iterator
+   * age for 3 consecutive minutes. Idle functions emit no datapoints and stay
+   * OK (`NOT_BREACHING`).
+   */
+  eventSourceIteratorAge: {
+    threshold: 60_000,
+    evaluationPeriods: 3,
+    datapointsToAlarm: 3,
+    treatMissingData: TreatMissingData.NOT_BREACHING,
```
laazyj, line 91: What was the authoritative source for these default alarm values? please include a link if this is a high-quality source.

laazyj (commented):
Minor documentation changes

packages/lambda/src/alarm-defaults.ts
laazyj: Good catch — there isn't an authoritative numeric source. AWS lists `IteratorAge` as a recommended metric for stream consumers but does **not** prescribe a threshold; it's explicitly workload-dependent (depends on your throughput and acceptable lag). My earlier `@see` pointed at the Recommended Alarms page, which covers Errors/Throttles/Duration/ConcurrentExecutions but not an IteratorAge threshold — so that citation was misleading.

Fixed in `71db7cf`: the `60_000 ms` / 3-datapoint values are now documented as a deliberately conservative, tunable default rather than an AWS-sourced number, and the `@see` now points to the [Lambda metrics doc](https://docs.aws.amazon.com/lambda/latest/dg/monitoring-metrics.html). The README gained a footnote saying the same. Values are unchanged and overridable via `eventSourceIteratorAge`.

(Also addressed the positional-"alarms above" comment in the same push by linking the two per-mapping fields explicitly.)

---

Issue #34, opened 2026-04-18
Add @composurecdk/sqs with QueueBuilder, DlqQueueBuilder, and SNS subscription DLQ auto-creation


laazyj:
Follow-up from #19 ([review thread](https://github.com/laazyj/composureCDK/pull/19#discussion_r3104983581)).

The AWS Well-Architected Reliability Pillar recommends attaching a DLQ to every SNS subscription. `SubscriptionBuilder` currently accepts a caller-owned `IQueue` via `.deadLetterQueue(...)` but does not create one automatically, because composureCDK has no SQS package yet and the queue is a caller-owned resource.

## Scope

1. **`@composurecdk/sqs` package** with a `createQueueBuilder()` that produces a Well-Architected, secure queue. Defaults worth considering:
   - `enforceSSL: true` (mirrors `TopicBuilder`).
   - KMS-managed server-side encryption.
   - A sensible retention period.
   - Standard recommended alarms (`ApproximateAgeOfOldestMessage`, `ApproximateNumberOfMessagesVisible`, `NumberOfMessagesSent == 0` staleness, etc. — these are queue-level and will be separate from the SNS topic-level redrive alarms added in #19).

2. **DLQ variant** — either a `createDlqQueueBuilder()` or a role/tag on `QueueBuilder` that applies DLQ-appropriate defaults and the DLQ-specific recommended alarms (different threshold profile — a DLQ with messages is itself an alert, whereas a primary queue with messages is normal).

3. **`SubscriptionBuilder` integration** — when no `deadLetterQueue` is provided, auto-materialise a DLQ using the new builder, with an opt-out for callers who don't want one. Should compose cleanly via `ref()` so callers can still pass their own queue.

4. **Resource policy wiring** — SNS needs permission to `sqs:SendMessage` on the DLQ (per [SNS DLQ docs](https://docs.aws.amazon.com/sns/latest/dg/sns-dead-letter-queues.html)). Make sure whichever builder owns the auto-DLQ attaches the right resource policy automatically.

## Rationale

- Brings SNS subscriptions closer to secure-by-default rather than secure-if-you-remember.
- Creates the foundational SQS package we'll need anyway for standalone queue builders.
- The DLQ-specific alarm distinction ([SNS redrive metrics are topic-level](https://github.com/laazyj/composureCDK/pull/19#discussion_r3104971760), DLQ health metrics are queue-level) falls out cleanly once the queue builder exists.

## Out of scope

[...]

PR #253, opened 2026-07-01
feat(sqs): add dead-letter-queue role switch to QueueBuilder

laazyj:
## What & why

Implements the DLQ-variant scope item from #34 using **Option B** from the [design discussion](https://github.com/laazyj/composureCDK/issues/34#issuecomment-4448537437) — a role flag on the existing `QueueBuilder` rather than a separate `createDlqQueueBuilder()`. One builder, one entry point, and it extends naturally to future roles (e.g. FIFO) without a new entry point.

`createQueueBuilder().asDeadLetterQueue()` switches the builder into the dead-letter-queue role:

- **Well-Architected queue defaults**: layers `DLQ_QUEUE_DEFAULTS` (14-day retention — the SQS maximum, maximizing the investigation/redrive window) on top of the existing secure defaults (`enforceSSL`, SSE-SQS, long polling).
- **Inverted recommended alarms**: any message present on a DLQ is itself the alert, so `approximateNumberOfMessagesVisible` (threshold > 0) is enabled by default, while the primary-queue "consumer falling behind" (`approximateAgeOfOldestMessage`) and "in-flight quota" (`approximateNumberOfMessagesNotVisible`) alarms — meaningless on a queue nothing actively consumes from — are off by default. Every alarm remains individually overridable via the existing `recommendedAlarms` API, in either direction.
- **Redrive-policy warning**: `Annotations`-based warning if a DLQ-role queue also configures its own `deadLetterQueue` redrive, since a DLQ is meant to be a terminal destination.

The role→enablement mapping is a single `Record<QueueRole, Record<QueueAlarmKey, boolean>>` table (`ALARM_DEFAULT_ENABLEMENT` in `queue-alarms.ts`), and alarm resolution is a single data-driven loop over `ALARM_SPECS` — adding a future role is additive (one more table row), not new branching.

This closes the DLQ-variant portion of #34. The remaining scope item — `SubscriptionBuilder` auto-creating a DLQ when none is provided — is left for a follow-up, since it would introduce a new `@composurecdk/sns` → `@composurecdk/sqs` package dependency (a first for this monorepo) and deserves its own discussion.

## New public API

- `createQueueBuilder().asDeadLetterQueue()`
- `DLQ_QUEUE_DEFAULTS`, `DLQ_ALARM_DEFAULTS` (exported for visibility/testing, per the existing `QUEUE_DEFAULTS`/`QUEUE_ALARM_DEFAULTS` convention)
- `QueueAlarmConfig` grows an `approximateNumberOfMessagesVisible` field (usable on primary queues too, as an opt-in with an explicit threshold)
- `QueueRole`, `QueueAlarmKey` types

## Checklist

- [x] Linked to an issue (#34)
- [x] Tests added/updated for the change (`packages/sqs/test/dlq-role.test.ts`, 51/51 sqs tests passing)

---
_Generated by [Claude Code](https://claude.ai/code/session_017Pudtcrpb5n5GQYzoM8gGg)_

### DIFF
--- a/packages/sqs/src/queue-builder.ts
+++ b/packages/sqs/src/queue-builder.ts
@@ -8,6 +8,8 @@ import { AlarmDefinitionBuilder } from "@composurecdk/cloudwatch";
 import type { QueueAlarmConfig } from "./queue-alarm-config.js";
 import { createQueueAlarms } from "./queue-alarms.js";
 import { QUEUE_DEFAULTS } from "./defaults.js";
+import { DLQ_QUEUE_DEFAULTS } from "./dlq-defaults.js";
+import type { QueueRole } from "./queue-role.js";
 
 /**
  * AWS-recommended minimum for `maxReceiveCount` on an SQS redrive
@@ -81,6 +83,13 @@ export interface QueueBuilderResult {
  * Alarms can be customized or disabled via the `recommendedAlarms` property.
  * Custom alarms can be added via the {@link addAlarm} method.
  *
+ * Calling `.asDeadLetterQueue()` switches the builder into the
+ * dead-letter-queue role: it applies {@link DLQ_QUEUE_DEFAULTS} (14-day
+ * retention) and inverts which recommended alarms are enabled by default
+ * — any message present becomes the alert, rather than a "consumer
+ * falling behind" or "in-flight quota" signal that doesn't apply to a
+ * queue nothing actively consumes from.
+ *
  * @see https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_sqs.Queue.html
  *
  * @example
@@ -88,6 +97,10 @@ export interface QueueBuilderResult {
  * const orders = createQueueBuilder()
  *   .queueName("orders")
  *   .visibilityTimeout(Duration.seconds(60));
+ *
+ * const ordersDlq = createQueueBuilder()
+ *   .queueName("orders-dlq")
+ *   .asDeadLetterQueue();
  * ```
  */
 export type IQueueBuilder = ITaggedBuilder<QueueBuilderProps, QueueBuilder>;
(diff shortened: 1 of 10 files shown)


packages/sqs/src/queue-builder.ts
```diff
@@ -184,3 +214,27 @@ function warnIfLowMaxReceiveCount(
...
+): void {
+  if (role !== "dlq") return;
+  if (!props.deadLetterQueue) return;
+  Annotations.of(scope).addWarningV2(
+    "@composurecdk/sqs:dlq-with-redrive-policy",
+    `QueueBuilder "${id}": built via asDeadLetterQueue() but also configures its own ` +
+      `deadLetterQueue redrive policy. A dead-letter queue is meant to be a terminal ` +
+      `destination for failed messages — redriving from it to another queue is unusual ` +
+      `and likely unintended.`,
+  );
```
laazyj, line 239: Ugh. This is probably the strongest argument against the "role" approach vs. specific builders. A shared property on the builder that has to be orphaned for the alternate role.

packages/sqs/src/queue-role.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * The recommended-defaults role a {@link createQueueBuilder | QueueBuilder}
+ * applies. `"primary"` is the default (a consumer-fed queue). `"dlq"`
+ * applies dead-letter-queue defaults and inverted alarm semantics via
+ * `.asDeadLetterQueue()`.
+ *
+ * Modeled as a union rather than a boolean so it extends to future roles
+ * (e.g. FIFO-tuned defaults) without another builder entry point.
+ */
+export type QueueRole = "primary" | "dlq";
```
laazyj, line 10: The crux of this decision is whether we can expand this role to include "fifo" - and I'm not sure if that's feasible.

laazyj (commented):
I'd like to see alternative approaches before making a decision

PR #115, opened 2026-05-13
feat(sqs): add @composurecdk/sqs with QueueBuilder

laazyj:
## Summary

Closes #112. First cut of `@composurecdk/sqs` — a fluent SQS queue builder with secure Well-Architected defaults and two AWS-recommended CloudWatch alarms. Mirrors the structure of `@composurecdk/sns` (`TopicBuilder`, `TopicAlarmConfig`, `TOPIC_ALARM_DEFAULTS`, `createTopicAlarms`).

Plan posted to the issue: https://github.com/laazyj/composureCDK/issues/112#issuecomment-4438621025

## What's in this PR

### Package

- `@composurecdk/sqs`, version `0.7.0`, peerDeps on `core` + `cloudformation` + `cloudwatch` + `aws-cdk-lib`.
- `createQueueBuilder()` — fluent builder, `taggedBuilder`-wrapped (Queue supports Tags), exposes every `QueueProps` field plus `recommendedAlarms` and `addAlarm`.
- `QueueBuilderResult = { queue: Queue, alarms: Record<string, Alarm> }`.
- `[COPY_STATE]` hook so `.copy()` preserves custom alarms (verified via `assertCopyPreservesState`).

### Secure defaults (`QUEUE_DEFAULTS`)

| Property                 | Default                       | Source                                                                                                                                                                                                                                                            |
| ------------------------ | ----------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `enforceSSL`             | `true`                        | [SNS/SQS security best practices](https://docs.aws.amazon.com/AmazonSQS/latest/SQSDeveloperGuide/sqs-security-best-practices.html) — deny on `aws:SecureTransport: false`. Mirrors `TopicBuilder`.                                                                |
| `encryption`             | `QueueEncryption.SQS_MANAGED` | [SQS data protection](https://docs.aws.amazon.com/AmazonSQS/latest/SQSDeveloperGuide/sqs-data-protection.html). KMS opt-in via `.encryption(QueueEncryption.KMS)` + `.encryptionMasterKey(key)`.                                                                  |
| `receiveMessageWaitTime` | `Duration.seconds(20)`        | [Long polling](https://docs.aws.amazon.com/AmazonSQS/latest/SQSDeveloperGuide/sqs-short-and-long-polling.html#sqs-long-polling) — fewer empty receives, lower cost, lower latency. 20s is the SQS maximum.                                                        |

[...]

PR #255, opened 2026-07-02
feat(sqs): add queue roles for FIFO and dead-letter queues (option B)

laazyj:
## What & why

Implements ergonomic FIFO (#117) and dead-letter (#34) queue support as **Option B** from the design discussions — a switchable queue role on a single builder — reworked to fix the awkwardness that stalled #253. This is one of **two competing PRs** — the sibling PR #254 implements Option A (sibling factories) with identical behaviour underneath, so the two API shapes can be compared directly.

### What was wrong with #253's role switch, and how this fixes it

#253's `.asDeadLetterQueue()` was a *fluent method*, and a fluent method can never narrow the builder's type: `IBuilder` regenerates the full prop surface from `Props` after every setter, so `deadLetterQueue` stayed callable on a DLQ and `fifo` stayed callable everywhere. The fix is to move role selection to the **factory argument**, where TypeScript overloads *can* bind a type: `createQueueBuilder(role?)` with `QueueRole = "standard" | "fifo" | "dlq" | "fifo-dlq"`, each role getting an exact prop surface from one builder implementation.

| Role | Surface changes | Behaviour |
| --- | --- | --- |
| `"standard"` (default) | FIFO-only props (`fifo`, `contentBasedDeduplication`, `deduplicationScope`, `fifoThroughputLimit`) removed from the type and rejected at build with a pointer to the FIFO roles | unchanged otherwise |
| `"fifo"` | `fifo` not settable (always `true`); `queueName` typed `` `${string}.fifo` `` so a missing suffix is a compile error | high-throughput mode without `deduplicationScope: MESSAGE_GROUP` throws; FIFO↔standard redrive-target mismatch throws; same secure defaults and primary alarm set — FIFO shares the standard 120k in-flight quota since Nov 2024, so no threshold fork (the "latent threshold bug" in #117 is moot; documented instead) |
| `"dlq"` | `deadLetterQueue` removed — a DLQ is terminal; FIFO-only props also absent | 14-day retention (`DLQ_QUEUE_DEFAULTS`); inverted alarms (below) |
| `"fifo-dlq"` | the FIFO surface minus `deadLetterQueue` | FIFO + DLQ behaviour combined — this is where Option B proves the use case: AWS requires a FIFO source's DLQ to be FIFO, and here that's **one more role, not a fourth factory or a leaked prop** |

The role is ordinary builder state (an internal prop), so `.copy()` preserves it for free — the gap #253 had to patch with `[COPY_STATE]` — and it's inspectable. Discovery is carried by the single entry point: autocomplete on the factory argument enumerates every role.

### DLQ alarm semantics (`DLQ_ALARM_DEFAULTS`)

- `approximateNumberOfMessagesVisible` **on**, threshold > 0 — any message on a DLQ is the alert.
- `approximateAgeOfOldestMessage` **on**, re-framed as "about to age out": default threshold is 75% of the queue's *resolved* `retentionPeriod` (`DLQ_AGE_ALARM_RETENTION_RATIO`), so tuning retention can't strand the alarm. Token-valued retention skips the alarm with the standardized acknowledgeable warning (`resolveAlarmThresholdBasis`, per #196).
- `approximateNumberOfMessagesNotVisible` **off** — nothing is in flight on a DLQ; opt back in via `recommendedAlarms`.

### Structure

- `queue-role.ts` — the `QueueRole` union + `isFifoRole`/`isDlqRole` predicates (roles are the fifo × dlq product).
- `queue-props.ts` — single home for the type carving: `FIFO_ONLY_PROP_KEYS`, `FifoQueueName`, per-role surfaces (the DLQ surfaces are structurally *derived* from their primary counterparts via `Omit<…, "deadLetterQueue">`, so they can't drift).
- `queue-validation.ts` — one `validateQueueProps(scope, id, role, props)` entry point; each guard owns its own applicability, so a new role never adds branching to the builder.
- `queue-alarms.ts` + `queue-alarm-profiles.ts` — one data-driven resolution loop over per-role profiles. An alarm with no generic baseline (visible-messages on a primary queue) **throws when enabled without an explicit threshold** instead of inheriting a noisy placeholder.

Full rationale in [ADR-0014](docs/adr/0014-role-parameterized-queue-builder.md) (status: Proposed — renumbered from 0013, which landed on `main` as consumer-side IAM grants while this PR was in review). PR #254 carries the competing Option A ADR.

### Breaking change

`QueueBuilderProps` no longer includes the FIFO passthrough props from #115; FIFO callers move to `createQueueBuilder("fifo")` unchanged otherwise. `IQueueBuilder` gains a defaulted type parameter (non-breaking for existing annotations).

Out of scope (unchanged from #253): `SubscriptionBuilder` auto-DLQ + SNS resource-policy wiring (#34 items 3–4).

## Checklist

- [x] Linked to an issue (#34, #117)
- [x] Tests added/updated for the change (75 sqs tests across all four roles, incl. validation throws, retention-scaled thresholds, token-retention skip, role-preserving copy semantics)

https://claude.ai/code/session_01DGnjGQrqGPaTL9mnfB5NKw


docs/adr/0013-role-parameterized-queue-builder.md
```diff
@@ -0,0 +1,107 @@
+# ADR 0013: Role-parameterized queue builder with per-role typed surfaces
```
laazyj: An ADR should be generalisable across multiple packages, otherwise it is just a package-scoped implementation detail (self documentation). 

There might be something here though. This could be a genuine architectural pattern to be used when CDK L2 API creates awkward mutually exclusive resource types. If that is true, then the ADR should describe the generalised pattern and when it applies (or doesn't) using QueueBuilder as the first example.

As per convention, the ADR will remain proposed until a second use case proves the decision.

docs/adr/0013-role-parameterized-queue-builder.md
```diff
@@ -0,0 +1,107 @@
...
+  so "FIFO dead-letter queue" is a real configuration, not an edge case.
+
+PR #253 prototyped a role switch as a fluent method (`.asDeadLetterQueue()`)
+on `createQueueBuilder()`. The behaviour was right, but the prop surface
+wasn't: every prop stayed reachable in every role — `deadLetterQueue` on a
+DLQ, `fifo` with primary-only alarm defaults — so invalid configurations
+were representable and only caught (if at all) at build time. The root cause
+is structural: `IBuilder<Props, T>` regenerates the full prop surface from
+`Props` after every setter, so a _mid-chain_ method can never narrow the
+type of what follows it.
```
laazyj: This should be framed as an alternative approach, with "alternatives" section at end of doc,  that was rejected because of the reasons stated.

docs/adr/0013-role-parameterized-queue-builder.md
```diff
@@ -0,0 +1,107 @@
...
+
+**Keep one entry point, but move role selection to the factory argument:
+`createQueueBuilder(role?)` with
+`QueueRole = "standard" | "fifo" | "dlq" | "fifo-dlq"`. Overloads bind each
+role to its own `Props` surface, so every role gets an exact fluent API from
+the same single builder implementation.**
+
+1. **The factory argument is the type boundary a fluent method can't be.**
+   `createQueueBuilder<R extends QueueRole>(role: R): IQueueBuilder<R>` with
+   `IQueueBuilder<R> = ITaggedBuilder<QueueBuilderPropsByRole[R], QueueBuilder>`.
```
laazyj: Unnecessary to include the ITaggedBuilder reference here, that is covered elsewhere. Keep the document focused on the key differentiating points of this decision

docs/adr/0013-role-parameterized-queue-builder.md
```diff
@@ -0,0 +1,107 @@
...
+   The per-role surfaces live in `queue-props.ts`: the standard surface omits
+   the FIFO-only props, the FIFO surfaces omit `fifo` (always `true`) and
+   retype `queueName` as `` `${string}.fifo` ``, and the dead-letter surfaces
+   omit `deadLetterQueue`. Untyped callers hit equivalent build-time guards
+   whose messages name the role to use instead.
+
+2. **The role is data, not class identity.** It is stored as an internal prop
+   (`queueRole`), set once by the factory and invisible on the public
+   surfaces. Because it rides in `props`, `.copy()` preserves it for free
+   (no `[COPY_STATE]` special case — the gap PR #253 had to patch), and it
```
laazyj: ADR shouldn't reference PRs that will be closed, it introduces draft concepts that were never part of the codebase

docs/adr/0013-role-parameterized-queue-builder.md
```diff
@@ -0,0 +1,107 @@
...
+   derives two booleans (`isFifoRole`, `isDlqRole`) and layers defaults
+   accordingly: `QUEUE_DEFAULTS` → `DLQ_QUEUE_DEFAULTS` (dlq roles) → user
+   props → `fifo: true` (fifo roles, after user props so it cannot be
+   unset).
+
+3. **Roles compose where factories multiply.** `"fifo-dlq"` demonstrates the
+   scaling argument: the types × roles product is four combinations today,
+   and each is one union member plus one row in the props map — not a new
+   entry point. The alternative (sibling factories, explored in a parallel
+   PR) needs either a factory per combination or a compromise where FIFO
```
laazyj: As above, don't reference "PR" but abstract alternatives to be noted at the end of the doc

laazyj (commented):
ADR needs work. Start there and I'll review rest of code.

docs/adr/0014-role-parameterized-queue-builder.md
```diff
@@ -0,0 +1,177 @@
...
+  "consumer falling behind" and "in-flight quota" signals do not apply.
+- The two axes **combine**: AWS requires a FIFO source's DLQ to itself be
+  FIFO, so a "FIFO dead-letter queue" is a real configuration, not an edge
+  case — four roles from a 2×2 product.
+
+The behaviour needed here is well understood; the open question is the API
+_shape_ that keeps each role's surface honest without multiplying entry
+points.
+
+## Decision
```
laazyj, line 49: This section covers the right points but goes too deep into the specific implementation details of the QueueBuilder. Raise the altitude to focus on the architecture and, optionally, link to the queue-builder.ts for easy discoverability of finer details

packages/sqs/src/index.ts
```diff
@@ -1,9 +1,20 @@
...
+  type FifoDlqQueueBuilderProps,
+  type QueueBuilderPropsByRole,
+} from "./queue-props.js";
 export { QUEUE_DEFAULTS } from "./defaults.js";
-export { type QueueAlarmConfig } from "./queue-alarm-config.js";
-export { QUEUE_ALARM_DEFAULTS } from "./queue-alarm-defaults.js";
+export { DLQ_QUEUE_DEFAULTS } from "./dlq-defaults.js";
+export { type QueueAlarmConfig, type QueueAlarmKey } from "./queue-alarm-config.js";
+export { QUEUE_ALARM_DEFAULTS, type QueueAlarmDefaults } from "./queue-alarm-defaults.js";
+export { DLQ_ALARM_DEFAULTS, DLQ_AGE_ALARM_RETENTION_RATIO } from "./dlq-alarm-defaults.js";
```
laazyj: Are all of these exports required? For example, the `DLQ_AGE_ALARM_RETENTION_RATIO` appears to be an internal implementation detail - external consumers could access it contextually via `DLQ_ALARM_DEFAULTS` member.

packages/sqs/src/queue-builder.ts
```diff
@@ -110,77 +107,80 @@ class QueueBuilder implements Lifecycle<QueueBuilderResult> {
...
-  Annotations.of(scope).addWarningV2(
-    "@composurecdk/sqs:redrive-low-max-receive-count",
-    `QueueBuilder "${id}": redrive policy maxReceiveCount is ${String(maxReceiveCount)}; ` +
-      `AWS recommends >= ${String(RECOMMENDED_MIN_MAX_RECEIVE_COUNT)} so the consumer ` +
-      `has room to retry before messages hit the dead-letter queue. ` +
-      `See https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html`,
-  );
+export function createQueueBuilder(): IQueueBuilder;
+export function createQueueBuilder<R extends QueueRole>(role: R): IQueueBuilder<R>;
+export function createQueueBuilder(role: QueueRole = "standard"): unknown {
```
laazyj, line 187: I'm concerned that by returning `unknown` this function breaks type safety in `compose`. Double-check that.

laazyj (commented):
Some further doc improvements and questions on the module's api surface

docs/adr/0014-role-parameterized-queue-builder.md
laazyj: Done in e0637a3. Rewrote the Decision section at a higher altitude — it now describes the four architectural parts of the pattern (role on the factory argument as the type boundary; role-as-data; roles composing as a product; per-role behaviour as data keyed by role) rather than QueueBuilder's specific prop surfaces, default-merge order, and alarm-profile internals, and links to `packages/sqs/src/queue-builder.ts` for those finer details.

---

packages/sqs/src/index.ts
laazyj: Good catch — dropped `DLQ_AGE_ALARM_RETENTION_RATIO` from the package barrel in e0637a3. It's an internal tuning constant: it stays exported from its own module for the internal profile code and the role tests (which import it via the relative path, not the barrel), and consumers can still reach the resolved value contextually through `DLQ_ALARM_DEFAULTS.approximateAgeOfOldestMessage.threshold`. The remaining barrel exports are the intended public surface — the `*_DEFAULTS` objects and the alarm-config types callers use to customise — so I left those as-is.

---

packages/sqs/src/queue-builder.ts
laazyj: Double-checked — the `unknown` doesn't reach `compose`. It's only the *implementation* signature; callers resolve the two typed overloads above it (`createQueueBuilder(): IQueueBuilder` and `createQueueBuilder<R>(role: R): IQueueBuilder<R>`), and TypeScript never exposes an overloaded function's implementation signature at a call site. So a builder passed to `compose` keeps its exact per-role type.

Verified it rather than reasoning about it — a throwaway type-level check against the package tsconfig:

```ts
const fifo: IQueueBuilder<"fifo"> = createQueueBuilder("fifo");   // compiles → not unknown
const asLifecycle: Lifecycle<QueueBuilderResult> = createQueueBuilder("fifo-dlq"); // flows into compose
fifo.contentBasedDeduplication(true);                            // ok on fifo
// @ts-expect-error not on the standard surface
createQueueBuilder().contentBasedDeduplication(true);
// @ts-expect-error not on the dlq surface
createQueueBuilder("dlq").deadLetterQueue(/* … */);
compose({ orders: createQueueBuilder().queueName("orders") }, { orders: [] }); // accepted, result type preserved
```

`tsc --noEmit` exits clean, which also confirms the three `@ts-expect-error` guards are real errors (an unused directive would fail the build). Added a short comment at the signature in e0637a3 so this doesn't resurface. Leaving the thread open for you to confirm the reasoning.

---

packages/sqs/src/index.ts
```diff
@@ -1,9 +1,20 @@
...
+  type FifoDlqQueueBuilderProps,
+  type QueueBuilderPropsByRole,
+} from "./queue-props.js";
 export { QUEUE_DEFAULTS } from "./defaults.js";
-export { type QueueAlarmConfig } from "./queue-alarm-config.js";
-export { QUEUE_ALARM_DEFAULTS } from "./queue-alarm-defaults.js";
+export { DLQ_QUEUE_DEFAULTS } from "./dlq-defaults.js";
+export { type QueueAlarmConfig, type QueueAlarmKey } from "./queue-alarm-config.js";
+export { QUEUE_ALARM_DEFAULTS, type QueueAlarmDefaults } from "./queue-alarm-defaults.js";
+export { DLQ_ALARM_DEFAULTS } from "./dlq-alarm-defaults.js";
```
laazyj, line 25: This is a huge increase in the api surface exported from this module - are they all required by consumers and/or architecture guidelines? Can some be trimmed as they are internal implementation detail only?

laazyj: Trimmed in e5dc87e. Went through the list export by export:

**Dropped** (internal plumbing, no consumer use, no emission obligation):
- `QueueBuilderExtensionProps` — heritage-only shared member; consumer `.d.ts` emission names the child interfaces, never the parent.
- `QueueAlarmKey` — derivable as `Exclude<keyof QueueAlarmConfig, "enabled">`; only the internal alarm loop uses it.
- `QueueAlarmDefaults` — the `*_ALARM_DEFAULTS` constants are declared with `satisfies`, so nothing public carries this alias as its type; `typeof QUEUE_ALARM_DEFAULTS` covers the rare consumer need.

**Kept**, each with a reason:
- `createQueueBuilder`, `IQueueBuilder`, `QueueRole`, `QueueBuilderResult` — the entry point, its parameter type, and its output type.
- The four per-role `*BuilderProps` + `QueueBuilderPropsByRole` — required by ADR-0001: after any setter call a consumer's inferred type expands to `ITaggedBuilder<FifoQueueBuilderProps, …>` (and generic helpers expand `IQueueBuilder<R>` to `ITaggedBuilder<QueueBuilderPropsByRole[R], …>`), so declaration emission must be able to name them or TS2883 fires — the exact 0.3.6 breakage that ADR records. Left a comment in the barrel so this doesn't get re-litigated.
- `FifoQueueName` — user-facing, not internal: it's the `queueName` parameter type on the FIFO surfaces, and callers who pre-declare names in typed variables/helpers need it by name.
- `QUEUE_DEFAULTS`, `DLQ_QUEUE_DEFAULTS`, `QUEUE_ALARM_DEFAULTS`, `DLQ_ALARM_DEFAULTS`, `QueueAlarmConfig` — the documented customisation surface, per architecture.md's "defaults are exported" convention; all appear in the README.

Also fixed the README, which still name-dropped the now-unexported `DLQ_AGE_ALARM_RETENTION_RATIO`.

Net: the barrel is main's surface plus `QueueRole`, `FifoQueueName`, the three new role props types + map, and the two DLQ defaults objects.

---

PR #256, opened 2026-07-02
test: enforce per-package coverage thresholds locally via vitest

laazyj:
## What & why

Following up on a discussion about whether Codecov would add value to this repo: line-coverage dashboards are easy to game here (a test can synthesise a stack without asserting on the output and still touch every line), and they're visibility-only unless paired with a required-status-check gate — nothing stops a regression on a contributor's own machine.

Instead, this wires up **branch-coverage enforcement that runs locally**, on the existing `test` target, so it's caught by `npm run verify` and the husky `pre-push` hook already in place — no new CI job, no hosted dashboard.

- `vitest.config.base.ts` (repo root) exports `withCoverage(thresholds, config?)`, wrapping `@vitest/coverage-v8` with `perFile: true` — an individual file below its package's threshold fails the build on its own, rather than being averaged into a package-wide number a single well-tested file could mask. That's what catches a brand-new builder/policy file shipped with no test.
- Every package's `vitest.config.ts` sets its own measured floor (from a baseline `vitest run --coverage` pass), since coverage genuinely varies by package — validation-heavy builder packages (`iam`, `route53`) sit well below mostly-declarative ones (`s3`, `sns`).
- Baselining surfaced three genuinely untested-but-testable gaps, closed with real tests rather than excluded or thresholded around:
  - `cloudwatch`'s `resolveAlarmConfig` had no direct test in its own package (only exercised indirectly via consuming packages).
  - `route53`'s `alias-targets.ts` helpers (CloudFront/API Gateway alias targets) had no `Ref`-branch coverage.
  - `events`' non-Lambda rule targets (SNS, SQS, EventBus, Step Functions, CloudWatch Logs) were missing `Ref`-branch tests, mirroring the existing `lambdaTarget` pattern.
- Ran this diff through a 4-angle simplify pass (reuse/simplification/efficiency/altitude) and applied the fixes: dropped a dead `overrides` parameter, reordered `withCoverage`'s signature so the 20 packages with no extra vitest config don't need to import `defineConfig` just to pass an empty object, and dropped the `html` coverage reporter (nothing consumes it; it wrote a full directory tree to disk on every run across 21 packages for no reader).
- Added [ADR-0013](docs/adr/0013-per-file-coverage-thresholds.md) documenting the decision, per AGENTS.md's requirement for new patterns.

## Checklist

- [x] Linked to an issue (or it's a small, obvious fix) — follow-up from conversation, not a filed issue
- [x] `npm run verify` passes locally
- [x] Tests added/updated for the change

[...]

PR #262, opened 2026-07-04
test(apigateway): reach 90% branch coverage and raise the floor

laazyj:
## What & why

Closes #257.

Raises `packages/apigateway`'s per-file branch-coverage floor from 75% to 90% (`vitest.config.ts`), backing it with tests for the four branches the #256 baseline left uncovered.

Two of the four were genuinely reachable and are closed with real tests:
- **`ResourceBuilder.applyTo`** — a method declared with no explicit integration (the `integration === undefined` arm) was never exercised; added a synth test for `addMethod("GET")` with no integration.
- **`resolveRestApiAlarmDefinitions`** — the "disable only one alarm" arm had no coverage for `latency`; extended the existing "disables individual alarms" test into a parameterized `it.each` covering `clientError` and `latency`, plus a direct unit test of the `enabled: false` guard, which `createRestApiAlarms` short-circuits before ever calling the helper.

The other two were unreachable defensive branches rather than test gaps, so they're removed rather than tested-around:
- `REST_API_DEFAULTS.deployOptions ?? {}` / `SPEC_REST_API_DEFAULTS.deployOptions ?? {}` could never take the `{}` arm — both constants always set `deployOptions`. Narrowing the constants with `satisfies Partial<…Props>` (instead of a `: Partial<…Props>` annotation that widened `deployOptions` back to optional) types `deployOptions` as always-present, so the fallback drops out. No behavioural change; the constant values are identical.

apigateway now reports 100% statements/branches/functions/lines.

## Checklist

### DIFF
--- a/packages/apigateway/src/defaults.ts
+++ b/packages/apigateway/src/defaults.ts
@@ -33,7 +33,7 @@ export const DEPLOY_OPTIONS_DEFAULTS: StageOptions = {
  * {@link createRestApiBuilder}. Each property can be individually overridden
  * via the builder's fluent API.
  */
-export const REST_API_DEFAULTS: Partial<RestApiBuilderProps> = {
+export const REST_API_DEFAULTS = {
   /**
    * Automatically create an access log group with structured JSON output.
    * Access logging provides an audit trail of all API calls for security
@@ -43,14 +43,14 @@ export const REST_API_DEFAULTS: Partial<RestApiBuilderProps> = {
   accessLogging: true,
 
   deployOptions: DEPLOY_OPTIONS_DEFAULTS,
-};
+} satisfies Partial<RestApiBuilderProps>;
 
 /**
  * Secure, AWS-recommended defaults applied to every spec-driven REST API
  * built with {@link createSpecRestApiBuilder}. Each property can be
  * individually overridden via the builder's fluent API.
  */
-export const SPEC_REST_API_DEFAULTS: Partial<SpecRestApiBuilderProps> = {
+export const SPEC_REST_API_DEFAULTS = {
   /**
    * Automatically create an access log group with structured JSON output.
    * @see https://docs.aws.amazon.com/wellarchitected/latest/serverless-applications-lens/opex-logging.html
@@ -58,4 +58,4 @@ export const SPEC_REST_API_DEFAULTS: Partial<SpecRestApiBuilderProps> = {
   accessLogging: true,
 
   deployOptions: DEPLOY_OPTIONS_DEFAULTS,
-};
+} satisfies Partial<SpecRestApiBuilderProps>;
--- a/packages/apigateway/test/rest-api-alarms.test.ts
+++ b/packages/apigateway/test/rest-api-alarms.test.ts
@@ -27,6 +27,8 @@ function mockIntegration() {
 
 const methodResponse200 = { methodResponses: [{ statusCode: "200" }] };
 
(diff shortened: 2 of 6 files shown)


packages/apigateway/src/defaults.ts
```diff
@@ -43,19 +43,19 @@ export const REST_API_DEFAULTS: Partial<RestApiBuilderProps> = {
   accessLogging: true,
 
   deployOptions: DEPLOY_OPTIONS_DEFAULTS,
-};
+} satisfies Partial<RestApiBuilderProps>;
```
laazyj, line 46: explain the thinking behind refactoring from a declared type to a "satisfies"?

packages/apigateway/test/rest-api-alarms.test.ts
```diff
@@ -203,6 +209,18 @@ describe("recommended alarms", () => {
       template.resourceCountIs("AWS::CloudWatch::Alarm", 2);
     });
 
+    it("disables only the latency alarm when set to false", () => {
```
laazyj: This is confusing as it doesn't describe what the subject is. It is also duplicative of the more generally described "disables individual alarms when set to false". Consider how you could make the existing "individual alarms" test parameterised to it tests the exclusive disablement of clientError & latency.

packages/apigateway/src/defaults.ts
laazyj: The goal was to remove the `?? {}` fallback on `REST_API_DEFAULTS.deployOptions` / `SPEC_REST_API_DEFAULTS.deployOptions` in the builders (an unreachable branch dragging coverage down), without touching the runtime values.

The original `: Partial<RestApiBuilderProps>` annotation widens every property — including `deployOptions` — to optional, even though the object literal always sets it. That optionality is what forced the `?? {}` at the call site.

`satisfies` checks the literal against `Partial<RestApiBuilderProps>` for compatibility but doesn't widen the inferred type, so `REST_API_DEFAULTS.deployOptions` stays typed as `StageOptions` (always present) rather than `StageOptions | undefined`. That lets the fallback come out cleanly, since it's now unreachable at the type level too, not just at runtime.

No behavioural change — same object, same values — just a narrower inferred type at the two call sites.

---

packages/apigateway/test/rest-api-alarms.test.ts
```diff
@@ -191,15 +197,19 @@ describe("recommended alarms", () => {
       template.resourceCountIs("AWS::CloudWatch::Alarm", 0);
     });
 
-    it("disables individual alarms when set to false", () => {
+    it.each([
+      { key: "clientError", others: ["serverError", "latency"] },
+      { key: "latency", others: ["clientError", "serverError"] },
+    ])("disables only the $key alarm when set to false", ({ key, others }) => {
```
laazyj: Why does the matrix not also test disabling "serverError"?

If this is duplicated by another test then fold it into here. Also consider a more efficient way to mange this test input matrix so it is easier to extend to other alarms in future. Perhaps "others" is an exclusion of key from a constant set?

laazyj: Fixed — pulled the matrix out into `ALARM_KEYS = ["clientError", "serverError", "latency"] as const` and derive `others` as `ALARM_KEYS.filter(other => other !== key)`, so `serverError` is now covered and any future alarm key gets exercised automatically without touching this test.

---

PR #267, opened 2026-07-04
test(ec2): reach 90/90/100 branch/statement/function coverage

laazyj:
## What & why

Closes #259.

Raises `packages/ec2`'s per-file floor from branches:66 / statements:84 / functions:75 to branches:90 / statements:90 / functions:100 (`vitest.config.ts`).

Five files were below the new floor:

- **`instance-alarms.ts`** / **`interface-endpoint-alarms.ts`** — both had a dead `config?.enabled === false` early return in their `resolve*Definitions` helper: `createInstanceAlarms`/`createInterfaceEndpointAlarms` already short-circuit on `enabled` before ever calling it (same unreachable-branch shape as apigateway's #257/#262). Removed rather than tested around. On the test side:
  - `instance-alarms.test.ts`: parameterized the existing single-alarm-disable test into an `it.each` covering `cpuUtilization`, `statusCheckFailed`, and `attachedEbsStatusCheckFailed` — only `cpuUtilization` had a disable test before.
  - `interface-endpoint-builder.test.ts`: added the missing `recommendedAlarms({ enabled: false })` test, and an `addAlarm` custom-alarm test (the custom-alarm resolve callback had no coverage at all).
- **`security-group-constraints.ts`** — `validateSecurityGroupDescription`/`validateSecurityGroupName`'s unresolved-token skip (ADR-0010) had no test in either validator. Added one per validator using a `CfnParameter`'s token-valued string.
- **`security-group-builder.ts`** — added a test for a validated `securityGroupName` (the `if (securityGroupName !== undefined)` branch had no test at all), and tests for `addEgressRule`/`addSelfIngress` with no description (the omitted-description ternary arm was untested; CDK synthesizes its own default description in that case, asserted directly).

`ec2` now reports 98.88% statements, 97.82% branches, 100% functions, with every file at or above the 90/90/100 floor.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01JMz3UmZvNhGpFf5KYq4XYX)_

### DIFF
--- a/packages/ec2/src/instance-alarms.ts
+++ b/packages/ec2/src/instance-alarms.ts
@@ -59,8 +59,6 @@ export function resolveInstanceAlarmDefinitions(
   config: InstanceAlarmConfig | undefined,
   props: Pick<InstanceProps, "instanceType">,
 ): AlarmDefinition[] {
-  if (config?.enabled === false) return [];
-
   const definitions: AlarmDefinition[] = [];
 
   if (config?.cpuUtilization !== false) {
--- a/packages/ec2/test/instance-alarms.test.ts
+++ b/packages/ec2/test/instance-alarms.test.ts
@@ -13,6 +13,12 @@ import {
 import { Metric } from "aws-cdk-lib/aws-cloudwatch";
 import { createInstanceBuilder } from "../src/instance-builder.js";
 
+const ALWAYS_ON_ALARM_KEYS = [
+  "cpuUtilization",
+  "statusCheckFailed",
+  "attachedEbsStatusCheckFailed",
+] as const;
+
 function buildInstance(
   configureFn?: (builder: ReturnType<typeof createInstanceBuilder>) => void,
   instanceType: InstanceType = InstanceType.of(InstanceClass.T3, InstanceSize.MICRO),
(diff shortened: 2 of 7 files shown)


packages/ec2/src/instance-alarms.ts
```diff
@@ -59,8 +59,6 @@ export function resolveInstanceAlarmDefinitions(
   config: InstanceAlarmConfig | undefined,
   props: Pick<InstanceProps, "instanceType">,
 ): AlarmDefinition[] {
-  if (config?.enabled === false) return [];
-
```
laazyj, line 63: Explain why these (this one and similar change in interface-endpoint-alarms.ts) input guards for the base config.enabled flag are not required. Is there a change in functionality here or is it a safe optimisation of an internal helper function?

packages/ec2/test/instance-alarms.test.ts
```diff
@@ -216,15 +216,28 @@ describe("recommended alarms", () => {
...
+      },
+      {
+        key: "statusCheckFailed",
+        others: ["cpuUtilization", "attachedEbsStatusCheckFailed", "cpuCreditBalance"],
+      },
+      {
+        key: "attachedEbsStatusCheckFailed",
+        others: ["cpuUtilization", "statusCheckFailed", "cpuCreditBalance"],
+      },
+    ])("disables only the $key alarm when set to false", ({ key, others }) => {
```
laazyj: Apply the same optimisation to this parameterised tests as was done in apigateway package (`ALARM_KEYS`).

packages/ec2/src/instance-alarms.ts
laazyj: No functionality change — it's the same unreachable-guard shape as apigateway's `resolveRestApiAlarmDefinitions` (#262).

`resolveInstanceAlarmDefinitions`/`resolveEndpointAlarmDefinitions` are only ever called from `createInstanceAlarms`/`createInterfaceEndpointAlarms`, and both of those already resolve `enabled` (defaulting from `INSTANCE_ALARM_DEFAULTS.enabled`/`INTERFACE_ENDPOINT_ALARM_DEFAULTS.enabled`) and early-return `{}` *before* calling the resolve helper:

```ts
const enabled = config?.enabled ?? INSTANCE_ALARM_DEFAULTS.enabled;
if (!enabled) return {};

const recommended = resolveInstanceAlarmDefinitions(instance, config, props);
```

So by the time `resolveInstanceAlarmDefinitions` runs, `config?.enabled === false` can never be true — the internal `if (config?.enabled === false) return [];` was dead. Neither helper is re-exported from the package's `index.ts` (only imported directly by whitebox tests), so there's no public entry point that could reach the guard with `enabled: false` and observe different behavior. Removed rather than tested-around, same as the apigateway precedent.

---

packages/ec2/test/instance-alarms.test.ts
laazyj: Fixed — pulled `ALWAYS_ON_ALARM_KEYS = ["cpuUtilization", "statusCheckFailed", "attachedEbsStatusCheckFailed"] as const` and derive `others` by excluding `key` from that set (`cpuCreditBalance` stays appended separately since it's contextual/burstable-only and already has its own dedicated test). Same approach as apigateway's `ALARM_KEYS`.

---

PR #282, opened 2026-07-05
feat(core): add combine() multi-ref combinator (#270)

laazyj:
Closes #270. Chosen over the option B builder in #276.

## What

Adds `combine(refs, transform?)` to `@composurecdk/core` — a **Ref combinator**. It resolves a record of `Resolvable` values against the build context and returns a **single `Ref`** to the merged record, optionally transformed:

```ts
addMethod("GET", combine(
  { table: ref<TableV2BuilderResult>("table"), role: ref<RoleBuilderResult>("apiRole") },
  ({ table, role }) => new AwsIntegration({
    service: "dynamodb", action: "Scan",
    options: {
      credentialsRole: role.role,
      requestTemplates: { "application/json": scanTemplate(table.table.tableName) },
    },
  }),
));
```

## Why

`ref()` reaches exactly one component, so a consumer that assembles a construct from **more than one** sibling — e.g. a direct API Gateway → AWS-service `AwsIntegration` needing both a target's identifier (for its VTL template) **and** a credentials role — had no single reference to hand to a `Resolvable<T>` seam. The workaround was a fake `Lifecycle` whose only job was to merge two sibling refs — a `compose` graph node modelling no resource, wired with untyped context casts.

`combine` closes that gap without new machinery:

- **Returns a `Ref`** — drops into every existing `Resolvable<T>` seam (`addMethod`, …) with no change to any consuming builder, and composes further with `.get()`/`.map()`.
- **Owns nothing** — creates no construct, holds no state; every sibling it references stays a first-class node in the `compose` graph.
- **Infers its types** — `Resolved<R>` mirrors `resolve()` at runtime (a `Ref<T>` yields `T`, a concrete value passes through), so the merged record's keys/types follow from the input and entries may mix refs and concrete values.

## Why not option B (#276)

The [#276](https://github.com/laazyj/composureCDK/pull/276) builder's real payload was auto-creating an IAM credentials role per integration — net-new machinery `aws-cdk-lib` does not provide, obscuring a part of the design CDK keeps explicit. `combine` addresses the actual, general gap instead; the credentials role stays an **explicit sibling** granted against with consumer-side grants ([ADR-0013](../blob/main/docs/adr/0013-consumer-side-grants.md)). AwsIntegration is just the motivating example — `combine` applies wherever one construct is assembled from several siblings.

## Docs & ADR

- `docs/architecture.md` — new "Combining refs — one consumer, many dependencies" subsection under **Ref**, with the example and explicit *when-not-to-use* guidance.
- **ADR-0015** — generalised decision: intent, when to use, when not to (single sibling → plain `ref`; permission wiring → consumer-side grants; not an ownership escape hatch). Indexed and cross-referenced from ADR-0013's out-of-scope note.

## Scope

Core + docs only, one cohesive change. No apigateway surface change, no auto-created roles. Option B's branch/PR #276 is left untouched to close separately.

## Testing

7 new unit tests (merge, independent resolution, transform, mixed concrete/ref entries, error propagation, `.map()`/`.get()` composition, `Resolvable`-seam interchangeability). 100% function/line coverage on `ref.ts`; `npm run verify` green across all 21 projects.

### DIFF
--- a/docs/adr/0015-combine-multi-ref-combinator.md
+++ b/docs/adr/0015-combine-multi-ref-combinator.md
@@ -1,0 +1,129 @@
+# ADR 0015: Combining multiple refs into one — the `combine()` Ref combinator
+
+- **Status:** Accepted
+- **Date:** 2026-07-05
+
+## Context
+
+`ref()` captures a lazy reference to **one** component's build output and resolves
+it against the build context during `build()`. A `Resolvable<T>` — the union
+`T | Ref<T>` — is the currency every builder seam speaks: `addMethod`,
+`addIntegration`, and the like all accept one, and internally call
+`resolve(value, context)`. This one-reference-per-value shape covers almost all
+wiring, because a consumer usually adapts a single dependency into the form it
+needs.
+
+Some constructs, though, are assembled from **more than one** sibling at once. A
+direct API Gateway → AWS-service `AwsIntegration` (API Gateway calling DynamoDB,
+SQS, S3, … without a Lambda hop) needs two siblings simultaneously: the target's
+identifier — a table's `tableName`, interpolated into the VTL request template —
+**and** the `credentialsRole` API Gateway assumes to make the call. A single
+`ref` reaches only one of them.
+
+Absent a combinator, the workaround was to introduce a component whose only job
+is to merge two sibling refs into one object — a fake `Lifecycle` that builds
+nothing:
+
+```ts
+// ANTI-PATTERN: a component that exists only to merge two sibling refs
+const tableAndRole = {
+  build: (_s, _i, ctx) => ({
+    tableName: (ctx.table as TableV2BuilderResult).table.tableName,
+    role: (ctx.apiRole as RoleBuilderResult).role,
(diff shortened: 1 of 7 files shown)


docs/adr/0015-combine-multi-ref-combinator.md
```diff
@@ -0,0 +1,139 @@
...
+    role: (ctx.apiRole as RoleBuilderResult).role,
+  }),
+};
+```
+
+This is a smell. It manufactures a node in the `compose` graph that models no
+resource, forces untyped `ctx[...] as ...` casts, and exists purely to work
+around `ref`'s single-component reach.
+
+An [earlier proposal](https://github.com/laazyj/composureCDK/pull/276) instead
```
laazyj: This paragraph doesn't need to be in the context as it needlessly distracts from the architecture decision. Expand the related "Alternatives considered" to include a little more detail - but not everything, still keep it short.

laazyj (commented):
Shorten the ADR to make it tighter

PR #287, opened 2026-07-06
feat(iam): OIDC identity-provider support with GitHub Actions layer

laazyj:
## Summary

Adds OIDC identity-provider support to `@composurecdk/iam`, resolving #278. The
request was GitHub-Actions-shaped, but the reusable primitive isn't — so this
ships as **two layers**: a general-purpose OIDC foundation (a legitimate IAM
concern) and a thin, opt-in GitHub Actions convenience layer built entirely on
top of it. This mirrors how `createServiceRoleBuilder` sits on `createRoleBuilder`.

Today consumers hand-roll the security-critical trust policy with raw
`aws-cdk-lib` — `ukehoot.net` and `jasonduffett.net` carry a byte-identical
`addCiOidc()`, and this repo's own CI solves it a third way in raw
CloudFormation. This collapses that to a small builder composition.

Full design rationale, alternatives considered, and the doctrine it's anchored
to (README "batteries included but overridable", ADR-0013, the defaults
doctrine): https://github.com/laazyj/composureCDK/issues/278#issuecomment-4890645464

## What's included

- **`createRoleBuilder().assumedBy(...)` now accepts `Resolvable<IPrincipal>`** —
  mirrors the existing `permissionsBoundary` wiring, so a principal derived from
  a provider built by a sibling `compose()` component can be referenced via
  `ref()` and resolved at build time. Purely additive.
- **General primitives:**
  - `createOidcProviderBuilder()` — fluent builder over the L2
    `OpenIdConnectProvider` (auto-fetches the root-CA thumbprint), validating a
    `https://` issuer URL.
  - `webIdentityPrincipal()` — builds an `OpenIdConnectPrincipal`, prefixing each
    bare claim key with the issuer host and mapping conditions onto
    `StringEquals`/`StringLike`. Removes the two error-prone steps of hand-writing
    the issuer prefix and choosing the operator.
- **GitHub Actions layer:**
  - `githubActionsProvider()` — provider builder preset with the GitHub issuer
    URL and `sts.amazonaws.com` audience, still fully overridable.
  - `githubActionsPrincipal()` — pins `aud` with `StringEquals` and scopes `sub`
    with `StringLike` so the operators can't be confused; **requires a non-empty
    subject list**, since a GitHub principal with no `sub` condition trusts every
    repository on GitHub.
  - `GitHubSubject` — type-safe `sub`-claim constructors (`branch`, `tag`,
    `pullRequest`, `environment`, `custom`) covering all three real-world shapes.
  - `importGithubActionsProvider()` — references the account-singleton provider
    by its conventional ARN.

## Design decisions

- **L2 `OpenIdConnectProvider`, not `OidcProviderNative`** — the native construct
  requires a thumbprint (breaks "doing nothing is secure/easy"); the L2
  auto-fetches it, matches every current consumer, and thumbprints are ignored
  for GitHub anyway since July 2023.
- **Trust policy delegated to CDK's `OpenIdConnectPrincipal`** — no hand-assembled
  IAM JSON (ADR-0013's "defer to the construct's own authority").
- **Secure defaults exported and Well-Architected-cited** per the defaults doctrine.
- **No new package** — the GitHub layer depends only on iam primitives; extraction
  to a `@composurecdk/ci` package is deferred until a second CI provider appears.

## Deferred

`createGithubActionsDeployRoleBuilder()` (the issue's optional "Layer 2") is
intentionally **not** included — its value versus how CDK-bootstrap-specific it is
is uncertain, and we'll revisit once Layers A–C are in use.

## Testing

- 45 tests across the iam package, 100% statement/branch/function/line coverage.
- `check:exports` (attw + publint) clean for the new barrel exports across
  CJS/ESM/bundler.
- `lint` and `format:check` clean.

Closes #278

### DIFF
--- a/packages/iam/src/index.ts
+++ b/packages/iam/src/index.ts
@@ -27,3 +27,4 @@ export {
   openIdConnectPrincipal,
   type OpenIdConnectPrincipalOptions,
 } from "./open-id-connect-principal.js";
+export * as githubActions from "./github-actions.js";
--- a/packages/iam/src/index.ts
+++ b/packages/iam/src/index.ts
@@ -17,3 +17,13 @@ export {
   StatementBuilder,
   WildcardResourceError,
 } from "./statement-builder.js";
+export {
+  createOpenIdConnectProviderBuilder,
+  type IOpenIdConnectProviderBuilder,
+  type OpenIdConnectProviderBuilderProps,
+  type OpenIdConnectProviderBuilderResult,
+} from "./open-id-connect-provider-builder.js";
+export {
+  openIdConnectPrincipal,
+  type OpenIdConnectPrincipalOptions,
+} from "./open-id-connect-principal.js";
--- a/packages/iam/README.md
+++ b/packages/iam/README.md
@@ -23,7 +23,7 @@ const role = createRoleBuilder()
   .build(stack, "StopEC2Role");
 ```
 
-Every [RoleProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_iam.RoleProps.html) property is available as a fluent setter. `permissionsBoundary` additionally accepts a `Resolvable<IManagedPolicy>` so a sibling component can supply a boundary policy via `ref(...)`.
+Every [RoleProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_iam.RoleProps.html) property is available as a fluent setter. `assumedBy` and `permissionsBoundary` additionally accept a `Resolvable` (`Resolvable<IPrincipal>` and `Resolvable<IManagedPolicy>`), so a sibling component can supply the trust principal or boundary policy via `ref(...)`.
 
 ### Defaults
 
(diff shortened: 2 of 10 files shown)


packages/iam/src/index.ts
```diff
@@ -17,3 +17,24 @@ export {
   StatementBuilder,
   WildcardResourceError,
 } from "./statement-builder.js";
+export {
+  createOidcProviderBuilder,
+  type IOidcProviderBuilder,
+  type OidcProviderBuilderProps,
+  type OidcProviderBuilderResult,
+} from "./oidc-provider-builder.js";
```
laazyj: This builder, and its supporting components, should mirror the L2 CDK construct that they provide (`OpenIdConnectProvider`). This is important discoverability for CDK experienced developers picking up this library.

This implies: `createOpenIdConnectProviderBuilder`, `IOpenIdConnectProviderBuilderProps`, open-id-connect-provider-builder.js`, etc.

packages/iam/src/index.ts
```diff
@@ -17,3 +17,24 @@ export {
...
+  GITHUB_OIDC_AUDIENCE,
+  GITHUB_OIDC_ISSUER_HOST,
+  GITHUB_OIDC_URL,
+  GitHubSubject,
+  type GitHubSubjectFactory,
+  githubActionsPrincipal,
+  githubActionsProvider,
+  type GithubActionsPrincipalOptions,
+  importGithubActionsProvider,
+} from "./github-actions.js";
```
laazyj: I want to expose GitHub helpers under a sub-namespace to isolate them from the standard `iam` API.

packages/iam/src/index.ts
```diff
@@ -17,3 +17,24 @@ export {
...
+  type OidcProviderBuilderResult,
+} from "./oidc-provider-builder.js";
+export {
+  webIdentityPrincipal,
+  type WebIdentityPrincipalOptions,
+} from "./web-identity-principal.js";
+export {
+  GITHUB_OIDC_AUDIENCE,
+  GITHUB_OIDC_ISSUER_HOST,
+  GITHUB_OIDC_URL,
```
laazyj: There is too much internal GitHub implementation detail exposed from the module. We do not want consumers to rely on us for these values, and I'm not sure that all the exported functions are part of the API vs. internal implementation details.

Review and consider each. But definitely don't export the constants.

laazyj (commented):
Pretty good, but tidy up naming and public API.

PR #294, opened 2026-07-11
feat(custom-resources): add AwsCustomResource builder

laazyj:
## What & why

Adds `@composurecdk/custom-resources`, exposing `createAwsCustomResourceBuilder()` — a compose-native wrapper around CDK's `AwsCustomResource` for AWS operations that have **no CloudFormation resource** (account-level SDK calls such as `ses:SetActiveReceiptRuleSet`, the call that makes an SES receipt rule set actually receive mail).

Using raw `AwsCustomResource` re-derives, every time: the lifecycle SDK calls, a `physicalResourceId`, the IAM policy, the `installLatestAwsSdk` choice, and a hand-written `node.addDependency(...)` — because the call sits outside the `compose` graph. This builder makes it a first-class citizen.

Resolves #280. Unblocks the SES `.activate()` in #279, which will consume this primitive.

## Usage

```ts
createAwsCustomResourceBuilder()
  .onUpdate({
    service: "SES",
    action: "setActiveReceiptRuleSet",
    parameters: ref<ReceiptRuleSetBuilderResult>("ruleSet").map((r) => ({ RuleSetName: r.ruleSet.receiptRuleSetName })),
    physicalResourceId: PhysicalResourceId.of("active-rule-set"),
  })
  .onDelete({ service: "SES", action: "setActiveReceiptRuleSet", parameters: {} })
  .dependsOn(ref<ReceiptRuleSetBuilderResult>("ruleSet"))
  .allow(["ses:SetActiveReceiptRuleSet"], ["*"]);
```

## Record of decisions

Durable notes on the non-obvious choices (an ADR was considered and dropped — this change justified the package but didn't establish a pattern that drives future behaviour; #279's domain-action work is where that ADR belongs, noted on that issue).

- **Dedicated `@composurecdk/custom-resources` package, not `core`.** `core` is `aws-cdk-lib`-free; this needs `aws-cdk-lib/custom-resources`. It stays idiomatic and discoverable via the same convention as other `aws-cdk-lib/*` wrappers.
- **`.dependsOn(...refs)` — an explicit, *precise* ordering seam.** `compose` decides build order, not deploy order; tokens buried in `AwsCustomResource`'s JSON-stringified `parameters` frequently don't produce a CFN `DependsOn` (why raw consumers hand-write `addDependency`). `.dependsOn(ref("x"))` adds a `DependsOn` to exactly the named component's construct(s), even when parameters are hardcoded strings. An **automatic walk of the whole build context was rejected** as too blunt — it couples the resource to incidental constructs and over-reaches into nested `parentContext`. (A precise *ref-provenance* auto-dependency remains a possible future enhancement if `.dependsOn` proves a common footgun.)
- **`.allow(actions, resources)` with `resources` required.** Sugar over `AwsCustomResourcePolicy.fromStatements`; requiring `resources` keeps a wildcard grant visible in review. `.policy(...)` is the full-control escape; supplying a `role` makes the policy optional.
- **Props derive from CDK:** `AwsCustomResourceBuilderProps = Omit<AwsCustomResourceProps, "onCreate" | "onUpdate" | "onDelete" | "policy">`, so the passthrough surface tracks the consumer's installed aws-cdk-lib automatically (later-introduced props appear with no maintenance; absent on older versions).
- **Not a tagged builder** — `Custom::AWS` has no `Tags`; uses plain `Builder` with the `builder-must-be-tagged` eslint-disable, as `@composurecdk/budgets` does.
- **Single default:** `installLatestAwsSdk: false` (determinism/speed); no invented resource-style defaults.

## Compatibility

`aws-cdk-lib` floor registered at **`2.37.0`** in `cdk-floors.json` (ADR-0008). Bisected via the `aws-custom-resource` `.d.ts`: the sole compile gate is `AwsCustomResourceProps.policy` becoming optional in 2.37.0 (required in 2.36.0), which the builder relies on to omit the policy when a role is supplied. Validated with `cdk-floors enforce 2.37.0` (build + typecheck + full suite against a real 2.37.0 install) and the CI floor shard.

## Tests & gate

- New package: synth assertions (lifecycle calls, IAM policy, `InstallLatestAwsSdk`), a precise `DependsOn` test (hardcoded parameter; asserts unnamed siblings are *not* depended on), Resolvable-parameter resolution, error paths, and an `assertCopyPreservesState` copy-state test. 100% coverage.
- Registered in `@composurecdk/module-compat` (ESM+CJS resolution passes).
- `npm run verify` green (format, build, catalogue, `attw`+`publint`, lint, cdk-floors, all tests); floor enforce green.


packages/custom-resources/package.json
```diff
@@ -0,0 +1,77 @@
...
+  "keywords": [
+    "aws",
+    "cdk",
+    "aws-cdk",
+    "infrastructure-as-code",
+    "iac",
+    "composurecdk",
+    "custom-resource",
+    "aws-custom-resource",
+    "escape-hatch"
```
laazyj: `escape-hatch` is not a relevant keyword. It's a terminology we've been using internally to describe a way to do things that our API surface doesn't naturally support.

Remove.

package-lock.json
```diff
@@ -9053,6 +9057,26 @@
...
+        "constructs": "^10.6.0",
+        "typescript": "^6.0.3",
+        "vitest": "^4.1.8"
+      },
+      "engines": {
+        "node": ">=20"
+      },
+      "peerDependencies": {
+        "@composurecdk/core": "^0.9.0",
+        "aws-cdk-lib": "^2.1.0",
```
laazyj: Should this be `^2.37.0`?

docs/adr/0016-generic-aws-custom-resource-escape-hatch.md
```diff
@@ -0,0 +1,119 @@
...
+on-delete SDK calls, a `physicalResourceId`, the IAM policy, the
+`installLatestAwsSdk` choice, **and** a hand-written `node.addDependency(...)` —
+because the call sits outside the `compose` graph and its parameters can't
+reference sibling components as `Ref`s.
+
+Every other ComposureCDK builder wraps _a resource_: secure defaults, recommended
+alarms, grants, a meaningful result. This one wraps _"call any SDK method"_
+(`service`/`action`/`parameters`). On the axes that justify most builders it is
+empty. That tension — a genuinely useful primitive that is nonetheless an escape
+hatch, not a resource abstraction — is what this ADR resolves.
```
laazyj: This is not sufficient for an ADR. It helps justify the specific package (`custom-resources`) vs. ignoring support for custom resources. But that is not novel as it remains aligned with `aws-cdk-lib/custom-resources` so is already idiomatic and discoverable via the same convention as other packages.

The "is not a resource" argument is also weak. There technically _is a resource_ (a lambda) that is created so having the `AwsCustomResource` appear in the compose graph is logical. 

Finally, the ADR doesn't drive future behaviour. It is purely a justification for this change. That is doesn't meet the bar for an ADR.

Capture the decision points in the PR description - this is a durable source of record that carries through the git log and release notes (changelog) for posterity.

Bottom line: drop ADR.

docs/adr/0016-generic-aws-custom-resource-escape-hatch.md
```diff
@@ -0,0 +1,119 @@
...
+   `installLatestAwsSdk: false` (determinism/speed) — no invented alarms or
+   resource-style defaults.
+
+5. **Prefer a domain builder.** For any _known_ call, a domain-specific method
+   (e.g. a future SES `.activate()`,
+   [#279](https://github.com/laazyj/composureCDK/issues/279)) is strictly better —
+   it scopes IAM automatically and reads as intent, never exposing
+   `service`/`action` strings. The generic builder is the floor for the long
+   tail, not the preferred path; its docs say so, mirroring ADR-0015's "when not
+   to use it".
```
laazyj: Once we implement #279 - that may justify an ADR as it will drive future behaviour as additional domain specific custom-resource behaviours are identified. Make a note on that issue.

packages/custom-resources/src/defaults.ts
```diff
@@ -0,0 +1,21 @@
+import { type AwsCustomResourceProps } from "aws-cdk-lib/custom-resources";
+
+/**
+ * Defaults applied to every custom resource built with
+ * {@link createAwsCustomResourceBuilder}. Each property can be individually
+ * overridden via the builder's fluent API.
+ *
+ * This is an escape-hatch builder, not a resource abstraction — it deliberately
+ * ships only the one default that is unambiguously safe. It does **not** invent
+ * recommended alarms, log groups, or other resource-style defaults.
```
laazyj: This paragraph (L8-10) is unnecessary. A user of custom resources will already understand.

packages/custom-resources/src/aws-custom-resource-builder.ts
```diff
@@ -0,0 +1,210 @@
...
+ * the lifecycle calls and policy, which the builder configures through its
+ * dedicated methods ({@link IAwsCustomResourceBuilder.onCreate | onCreate} /
+ * {@link IAwsCustomResourceBuilder.allow | allow} /
+ * {@link IAwsCustomResourceBuilder.policy | policy} /
+ * {@link IAwsCustomResourceBuilder.dependsOn | dependsOn}).
+ *
+ * Deriving from `AwsCustomResourceProps` means the fluent surface tracks the
+ * consumer's installed aws-cdk-lib: props added in later releases (e.g.
+ * `serviceTimeout`, `logGroup`, `memorySize`) are available automatically with
+ * no floor bump, and are simply absent on older versions.
```
laazyj: L22-25 is unnecessary - this is the convention followed throughout the project.

packages/custom-resources/src/aws-custom-resource-builder.ts
```diff
@@ -0,0 +1,210 @@
...
+ * blessed, compose-native escape hatch for AWS operations that have **no
+ * CloudFormation resource** (account-level SDK calls such as
+ * `ses:SetActiveReceiptRuleSet`).
+ *
+ * This is an escape hatch, not a resource abstraction. **When a domain builder
+ * exists for the call you need, prefer it** — it scopes IAM automatically and
+ * reads as intent rather than plumbing. Reach for this builder only for the
+ * long tail of one-off SDK calls that don't justify a domain builder.
+ *
+ * What it genuinely buys over raw `AwsCustomResource`:
```
laazyj: Phrasing on L58 is inconsistent with, too casual for, TSDoc describing a public interface. Try "Benefits over raw `AwsCustomResource`:"

packages/custom-resources/src/dependencies.ts
```diff
@@ -0,0 +1,55 @@
...
+ * constructs (`{ bucket }`), or nested maps of constructs (`{ alarms: {...} }`),
+ * so the walk descends into plain objects and arrays but **stops at the first
+ * construct** — its descendants are ordered transitively by CloudFormation, and
+ * descending into a construct's own `node` tree would pull in unrelated nodes.
+ *
+ * @param value - The value to search (typically a component's build result).
+ * @param out - Accumulates the constructs found.
+ * @param seen - Guards against cycles in plain-object graphs.
+ */
+export function collectConstructs(
```
laazyj, line 18: Is this exported entirely for testing? If so, annotate it so it is less likely to leak to the public api.

PR #301, opened 2026-07-12
feat(ses): receiving-side builders (new @composurecdk/ses)

laazyj:
## Summary

Adds `@composurecdk/ses` — the SES **receiving-side** cut requested in #279. Everything inbound is now compose-native instead of raw `aws-cdk-lib`.

- **EmailIdentity** — `.domain()` / `.email()` / `.publicHostedZone()`, Easy/BYO DKIM, MAIL FROM, and mode-aware `.publishDkim(zone)` (three CNAMEs for Easy DKIM, one TXT for BYODKIM) via route53's absolute-name primitive. Result exposes the DKIM tokens and the published records.
- **ReceiptRuleSet** — inner `.rule(key, cfg)` sub-builder (ordered), `dropSpam` passthrough, and opt-out `.activate()`.
- **ReceiptFilter** / **allow-list** — IP block / allow lists.
- **Actions** — free-function helpers `s3Action` / `lambdaAction` / `snsAction` / `bounceAction` / `stopAction` / `addHeaderAction`, all accepting `Resolvable` inputs so a rule wires to sibling components by `ref()`.

## Key decisions

- **Activation is on by default (opt-out) with conditional deactivation.** A rule set is inert until it is the account's active set — a footgun with no CFN resource. `.activate()` is backed by a **purpose-built provider** (not the single-call `AwsCustomResource`) whose delete handler clears the active slot *only if it is ours*, so teardown never disables another stack's rule set. New **[ADR-0016](../blob/feat/ses-receiving-builders/docs/adr/0016-domain-action-custom-resource.md)** captures this domain-action-over-custom-resource pattern (as flagged in the #280/#294 review).
- **`.publishDkim()` lives on the identity builder** (per the #281 amendment): the DKIM tokens are only concrete inside `build()`, and DSL record *names* aren't `Resolvable`. DKIM domain-knowledge (3 CNAMEs vs. 1 TXT) lives here; route53 owns only the DNS primitive. `.publicHostedZone()` stays for the apex-owns-zone case (CDK auto-publishes) and is mutually exclusive with `.publishDkim()`.
- **Props = `Omit<CdkProps, builder-owned keys>` + additions**, so new `aws-cdk-lib` props pass through (forward-compatible). `dropSpam` is a free passthrough.
- **Action helpers are free functions** (matching `events/targets`), pulling **zero** `@composurecdk` peers — they reference `aws-cdk-lib` interface types (`IBucket`/`IFunction`/`ITopic`/`IKey`). The package peers only `@composurecdk/core` and `@composurecdk/route53`; no cycles.

## Secure defaults (WAF Security Pillar)

| Default | Value | Note |
| --- | --- | --- |
| `scanEnabled` | `true` | Spam/virus scan (CFN default is off) |
| `tlsPolicy` | `REQUIRE` | Encrypt in transit. ~95%+ of legit inbound already uses TLS and a non-TLS sender gets a **hard bounce** (fail-loud); overridable via `.tlsPolicy(OPTIONAL)` |
| `mailFromBehaviorOnMxFailure` (when MAIL FROM set) | `REJECT_MESSAGE` | No insecure `amazonses.com` fallback; preserves DMARC alignment |
| `s3Action` | bucket policy + KMS grant | Encryption at rest works out of the box |

A synth-time warning fires when a receiving construct is built in a Region that can't receive mail.

## Test plan

- **100% coverage** (statements/branches/functions/lines) across all modules; 49 tests.
- Registered in `@composurecdk/module-compat` (`DUAL_PACKAGES` + peer) — dual ESM/CJS resolution passes.
- Added to `cdk-floors.json` (floor **2.216.0**, gated by the route53 peer's zone DSL); `cdk-floors:check` green.
- Full `npm run verify` passes locally (build, `check:exports`, lint, format, catalogue, floors, test).

## Not in this PR

Sending-side builders — verified send identity, configuration sets + event destinations, suppression list, VDM, dedicated IP pools, and reputation `recommendedAlarms` (account/config-set-scoped, **not** hung off the identity). Deferred to a follow-up issue.

Refs #279.


packages/ses/package.json
```diff
@@ -0,0 +1,83 @@
+{
+  "name": "@composurecdk/ses",
+  "version": "0.9.0",
+  "description": "Compose-native SES receiving-side builders — EmailIdentity, ReceiptRuleSet, ReceiptFilter, and rule actions",
```
laazyj: The description is unnecessarily narrow. the package _will_ support the full SES functionality in the near future - it just has receive-side for this commit.

Generalise the description around full SES support

packages/ses/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Install
+
+```sh
+npm install @composurecdk/ses
+```
+
+Peer dependencies: `@composurecdk/core`, `@composurecdk/route53` (for `.publishDkim()`), `aws-cdk-lib`, `constructs`.
+
+## Email identity
```
laazyj, line 15: Each component should have a 1-2 sentence explanation of what it is / what it does. It's ok if this borrows from the underlying CDK construct's description.

packages/ses/README.md
```diff
@@ -0,0 +1,91 @@
...
+          objectKeyPrefix: "inbound/",
+        }),
+      )
+      .addAction("process", lambdaAction(ref<FunctionBuilderResult>("intake").get("function"))),
+  )
+  .build(stack, "MailRuleSet");
+```
+
+- **Rules run in declaration order**, and actions run in the order added — ordering is semantic (a `Stop`/`Bounce` action halts later rules), so it is preserved.
+- **Actions are free functions** — `s3Action`, `lambdaAction`, `snsAction`, `bounceAction`, `stopAction`, `addHeaderAction` — each accepting `Resolvable`s so a rule wires to sibling components by `ref()`.
```
laazyj: Describe the underlying action argument. The free form functions are helpers - they are useful but optional (i.e. where we do not yet provide a helper for a specific action type).

packages/ses/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Activation (on by default)
+
+A receipt rule set is **inert until it is the account's active rule set** — a separate, account-level `ses:SetActiveReceiptRuleSet` call with no CloudFormation resource. Miss it and mail silently vanishes, so `.activate()` is **on by default** (opt out with `.activate(false)`).
+
+Activation is backed by a purpose-built provider (see [ADR-0016](../../docs/adr/0016-domain-action-custom-resource.md)) that **conditionally deactivates**: on delete it clears the active slot only if the currently-active set is this one, so tearing down a stack never disables another stack's rule set. The custom resource is exposed on the result as `activation`.
+
+> Only **one** rule set is active per account/region. If you run multiple rule sets across stacks and need bespoke arbitration, disable activation here and drive `setActiveReceiptRuleSet` yourself with [`@composurecdk/custom-resources`](../custom-resources).
```
laazyj, line 79: Link to AWS documentation explaining this limitation

packages/ses/src/actions.ts
```diff
@@ -0,0 +1,145 @@
+import { ServicePrincipal } from "aws-cdk-lib/aws-iam";
```
laazyj: Spilit the specific actions to their own files making `actions` a directory with a barrel.

packages/ses/src/actions.ts
```diff
@@ -0,0 +1,145 @@
...
+      ...(objectKeyPrefix !== undefined && { objectKeyPrefix }),
+      ...(resolved.kmsKey && { kmsKey: resolved.kmsKey }),
+      ...(resolved.topic && { topic: resolved.topic }),
+    });
+  });
+}
+
+/** Options for {@link lambdaAction}. */
+export interface LambdaActionOptions {
+  /** Whether SES invokes the function asynchronously (`Event`) or waits (`RequestResponse`). */
```
laazyj: Specify the default if `invocationType` is not supplied.

packages/ses/src/actions.ts
```diff
@@ -0,0 +1,145 @@
...
+        function: resolved.fn,
+        ...(invocationType !== undefined && { invocationType }),
+        ...(resolved.topic && { topic: resolved.topic }),
+      }),
+  );
+}
+
+/** Options for {@link snsAction}. */
+export interface SnsActionOptions {
+  /** Encoding SES uses for the message published to the topic. */
```
laazyj: Specify the default value if `encoding` is not provided.

packages/ses/src/activation.ts
```diff
@@ -0,0 +1,69 @@
...
+ * A rule set is inert until it is the account's single **active** rule set —
+ * `ses:SetActiveReceiptRuleSet`, which has no CloudFormation resource. On
+ * create/update it activates the named rule set. On delete it **conditionally
+ * deactivates**: it clears the active slot only when the currently-active set
+ * is the one being deleted, so tearing down this stack never disables another
+ * stack's rule set. (A single-call `AwsCustomResource` cannot express this
+ * describe-then-act logic, which is why activation uses a purpose-built
+ * provider.) `@aws-sdk/client-ses` is provided by the Lambda runtime.
+ */
+const HANDLER = `
```
laazyj, line 25: Could we expose the handler code in an actual JS file (CJS in this case) so that it benefits from static syntax checking?

packages/ses/src/activation.ts
```diff
@@ -0,0 +1,69 @@
...
+ * {@link IReceiptRuleSetBuilder.activate | `.activate()`}; returned on the build
+ * result so consumers can reference the custom resource.
+ */
+export function activateReceiptRuleSet(
+  scope: IConstruct,
+  id: string,
+  ruleSet: IReceiptRuleSet,
+): CustomResource {
+  const onEvent = new LambdaFunction(scope, `${id}Fn`, {
+    runtime: Runtime.NODEJS_20_X,
```
laazyj: Prefer a later LTS release so we have a longer window before this Lambda starts generating warnings about Node version deprecation. At this time NODEJS_24_X is the latest LTS version.

If this forces the supported CDK floor to rise then consider if a magic string can be used rather than relying on the `Runtime` enum.

packages/ses/src/activation.ts
```diff
@@ -0,0 +1,69 @@
...
+ * deactivates it on delete. Bundled by
+ * {@link IReceiptRuleSetBuilder.activate | `.activate()`}; returned on the build
+ * result so consumers can reference the custom resource.
+ */
+export function activateReceiptRuleSet(
+  scope: IConstruct,
+  id: string,
+  ruleSet: IReceiptRuleSet,
+): CustomResource {
+  const onEvent = new LambdaFunction(scope, `${id}Fn`, {
```
laazyj, line 52: Consider - if we already have the `@composurecdk/lambda` dependency, can we utilise `createFunctionBuilder()` - disabling any of the "batteries included" functionality we specifically don't want for a custom resource lambda.

packages/ses/src/activation.ts
```diff
@@ -0,0 +1,69 @@
...
+  // Account-level actions: SES rule-set activation is not resource-scoped.
+  onEvent.addToRolePolicy(
+    new PolicyStatement({
+      effect: Effect.ALLOW,
+      actions: ["ses:SetActiveReceiptRuleSet", "ses:DescribeActiveReceiptRuleSet"],
+      resources: ["*"],
+    }),
+  );
+  const provider = new Provider(scope, `${id}Provider`, { onEventHandler: onEvent });
+  return new CustomResource(scope, id, {
```
laazyj, line 67: Consider: can we utilise `@composurecdk/custom-resources` instead of raw CDK construct here? If not, what is missing from that API?

packages/ses/src/email-identity-builder.ts
```diff
@@ -0,0 +1,243 @@
...
+ * Configuration for the SES email-identity builder. The `identity` and
+ * `dkimIdentity` props are owned by the builder's fluent methods
+ * ({@link IEmailIdentityBuilder.domain | `.domain()`} etc.); every other
+ * {@link EmailIdentityProps} field passes through unchanged.
+ */
+export type EmailIdentityBuilderProps = Omit<EmailIdentityProps, "identity" | "dkimIdentity">;
+
+/** DKIM token names and values exposed on the build result. */
+export interface DkimTokens {
+  readonly tokenName1: string;
```
laazyj: This struct is bound to exactly 3 tokens, which matches the current implementation but is an arbitrary number.

Consider if this could be modelled as a simple read only map

packages/ses/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  bounceAction,
+  type BounceActionOptions,
+  stopAction,
+  addHeaderAction,
+} from "./actions.js";
+export { DEFAULT_RECEIPT_RULE, DEFAULT_MAIL_FROM_BEHAVIOR_ON_MX_FAILURE } from "./defaults.js";
+export {
+  SES_RECEIVING_REGIONS,
+  RECEIVING_REGION_WARNING,
+  warnIfNotReceivingRegion,
```
laazyj: This helper function shouldn't be exported as part of the package api

docs/adr/0016-domain-action-custom-resource.md
```diff
@@ -0,0 +1,95 @@
+# ADR 0016: Domain-action custom resources — purpose-built providers for stateful SDK-only operations
+
+- **Status:** Accepted
```
laazyj: As per convention, this will be Proposed state until a second instance is implemented to price the decision

docs/adr/0016-domain-action-custom-resource.md
```diff
@@ -0,0 +1,95 @@
...
+- **Date:** 2026-07-12
+
+## Context
```
laazyj, line 6: The decision is sound but the context leaves too heavily on the details of the ses example which makes it harder to discern the general applicability.

Rewrite the context either this in mind and try to shorten it

packages/ses/src/activation.ts
laazyj: Dug into this — the blocker is fundamental to `AwsCustomResource`, not something the builder can close.

`AwsCustomResourceBuilder` wraps `AwsCustomResource`, whose model is **one declarative SDK call per lifecycle event**: each of `onCreate`/`onUpdate`/`onDelete` is a single `SdkCallConfig` (`service` + `action` + synth-fixed `parameters`). Conditional deactivation needs this on delete:

```
active = SES.describeActiveReceiptRuleSet()        // call #1
if (active?.Metadata?.Name === ourRuleSetName)     // branch on the result
    SES.setActiveReceiptRuleSet({})                // call #2, only sometimes
```

That's three things `AwsCustomResource` structurally can't express, each independently fatal:

1. **Two SDK calls in one event** — `onDelete` is a single call.
2. **Branch on a call's response** — `parameters` are frozen at synth (or ref-resolved at synth), never computed from a runtime response.
3. **Read-modify-write on live remote state** — no "run this code" hook.

The escape hatches don't help either: two chained custom resources can't say "only run #2 if #1 returned X" (the second always fires), and `ignoreErrorCodesMatching` swallows errors rather than adding conditionality.

The only thing `AwsCustomResource` *can* do here is `onDelete → setActiveReceiptRuleSet({})` **unconditionally** — which is exactly the clobber-another-stack's-active-set footgun (the downstream-consumer review on this PR confirmed they hit it with their hand-rolled version).

So what's missing is a **user-code execution hook** — a handler that branches on remote state. That isn't a small addition to `AwsCustomResource`; it's a different construct (`Provider` + a handler Lambda), and bolting it onto a builder whose whole identity is "declarative single SDK call" would conflate two abstractions. Keeping the purpose-built provider is the right call — and it's why ADR-0016 is the seam. I'm moving that ADR to **Proposed**: if a second domain-action lands, that prices whether a shared handler/provider builder is worth extracting (possibly into `@composurecdk/custom-resources` itself).

packages/ses/src/activation.ts
laazyj: Compared the two paths — I'd keep the raw `Function` here.

First a correction to the premise: `@composurecdk/lambda` isn't currently a dependency of `ses`. The action helpers import the *interface* `IFunction` from `aws-cdk-lib/aws-lambda` (same as `events/targets/lambda-target.ts`), not the builder — so going through `createFunctionBuilder()` would **add a new `@composurecdk/lambda` peer** (plus its transitive `@composurecdk/cloudwatch`).

The provider handler is internal custom-resource plumbing, but the lambda builder's defaults are aimed at *user* functions: `tracing: ACTIVE`, `loggingFormat: JSON`, and recommended CloudWatch alarms on by default (its result also manages a `LogGroup`). To use it here I'd be writing `.tracing(DISABLED).recommendedAlarms(false)` — importing a batteries-included abstraction specifically to switch its batteries off, and still emitting an extra managed LogGroup on a throwaway Lambda.

Two things also blunt the consistency argument:

- The `Provider` framework spins its **own** raw Lambda (the onEvent/isComplete wrapper) that can't route through the builder anyway — so half the activation Lambda footprint stays raw regardless.
- `activation.ts` already reaches for raw `Provider` and `CustomResource` (neither has a builder), so "no raw L2 in a composure package" isn't an invariant this file can hold.

So raw `Function` is the right altitude — deep plumbing, the way CDK's own `Provider` uses it. Happy to revisit if we ever want a hard "never instantiate a raw L2" rule across the org, but that's a bigger call than this PR.

packages/ses/src/activation.ts
laazyj: Prototyped the shipped-`.cjs`-asset route and hit a wall in the dual-publish setup, so I've taken a different path that I think is actually better.

**Why not a shipped asset:** `tshy` doesn't copy non-TS files into `dist` (verified — the `.cjs` just doesn't appear in the build output). And even if it did, locating the asset from `src/` at synth time needs `__dirname` (undefined in the ESM build) or `import.meta.url` (banned by `no-cjs-incompatible-syntax` / ADR-0007), so there's no portable path across the ESM+CJS builds. Making it work would need a custom copy step plus a path shim — first-of-kind tooling here, and no other package ships a Lambda asset.

**What I did instead** (986916b): extracted the decision logic into a typed function `runActivation(event, api)` in `activation-handler.ts`, and serialise *just that* into the Lambda via `.toString()`, leaving only a thin SDK adapter inline:

```js
const { SESClient, SetActiveReceiptRuleSetCommand, DescribeActiveReceiptRuleSetCommand } = require("@aws-sdk/client-ses");
async function runActivation(event, api) { /* conditional-deactivate decision */ }
exports.handler = async (event) => {
  const ses = new SESClient({});
  const api = { getActiveRuleSetName: /* describe */, setActive: /* set/clear */ };
  return runActivation(event, api);
};
```

The `api` seam abstracts the two SES calls, so the conditional-deactivate logic is now **type-checked and unit-tested** — `activation-handler.test.ts` covers all four branches (create, update, delete-ours, delete-someone-else's). That also closes the "inline handler is untested" caveat ADR-0016 called out; only the ~5-line SDK adapter stays unchecked. No packaging change, no new deps, coverage stays 100%.

If you'd still prefer a genuinely shipped asset down the line, that's a separate tooling task (copy step + portable path helper) — happy to file it as a follow-up.

packages/ses/src/activation.ts
laazyj: Done in 986916b — moved to Node 24, using a magic-string runtime rather than the `Runtime.NODEJS_24_X` enum exactly as you suggested, so the LTS bump doesn't drag the aws-cdk-lib floor up to whatever release introduced the enum member:

```ts
const NODE_RUNTIME = new Runtime("nodejs24.x", RuntimeFamily.NODEJS, { supportsInlineCode: true });
```

Floor stays at 2.216.0 (route53-driven).

packages/ses/src/email-identity-builder.ts
laazyj: Agreed the fixed 3-token struct was arbitrary. Rather than a map, I've exposed CDK's own `dkimRecords` — `readonly { name: string; value: string }[]` — directly on the result as `dkim` (986916b). It's the natural non-fixed-arity model and matches what the L2 already surfaces, so there's a single source of truth. (A name→value map would key on unresolved tokens, which reads oddly; the array sidesteps that.)

docs/adr/0016-domain-action-custom-resource.md
```diff
@@ -0,0 +1,34 @@
...
+
+- **Status:** Proposed
+- **Date:** 2026-07-12
+
+## Context
+
+Some AWS operations have **no CloudFormation resource** — they are SDK calls only. [`@composurecdk/custom-resources`](../../packages/custom-resources) wraps CDK's `AwsCustomResource` for these, and for a **stateless** operation (one fixed SDK call per lifecycle event) it is the right tool.
+
+A subset are **stateful**: the correct action on a lifecycle event depends on current remote state. The delete handler in particular may need to _read_ state, _decide_, then _maybe act_ — a read-modify-write. `AwsCustomResource` runs one fixed call per event with parameters frozen at synth time and no hook to branch on a call's response, so it structurally cannot express this. Forcing such an operation through it means falling back to an unconditional call, which is often unsafe — e.g. clobbering account-level state another stack owns.
+
```
laazyj: I'm not liking the reframing - we're in danger of losing the decisions where a _stateless_ API call should be modelled as an operation on the builder to keep it encapsulated within, and using the language of, its domain. For example, if we'd decided to allow unconditional delete for the `activation` wrapper.

This context focuses too much on the stateful/stateless use cases and misses the important decision: To encapsulate AWS SDK API calls, that in a raw CDK system would be modelled as `AwsCustomResource` or a Provider, as operations on the builder that naturally owns the domain logic.

The stateful/stateless bi-modality is an implementation decision. The guidance provided on how to identify and implement these two modes is valuable, but it is irrelevant without the higher level architectural decision to encapsulate the custom resource within the builder.

Try again, considering this framing.

laazyj (commented):
ADR still doesn't land. See comment inline

docs/adr/0016-domain-action-custom-resource.md
laazyj: Good call — reframed in bee4ae8. The Context and Decision now lead with the architectural decision you name: **encapsulate an SDK-only operation (that raw CDK would model as a standalone `AwsCustomResource` / `Provider`) as a domain action — a `<verb>()` method — on the builder that owns its domain**, expressed in the domain's language and owning its IAM/lifecycle, with the custom resource surfaced on the build result.

The stateless-vs-stateful split is now explicitly demoted to *secondary implementation guidance* for how to back the operation, with a note that the same `.activate()` encapsulation would hold even if activation used an unconditional stateless `AwsCustomResource`. Retitled to "Encapsulate SDK-only operations as domain actions on the owning builder".

docs/adr/0016-domain-action-custom-resource.md
laazyj: Addressed together with your follow-up below (bee4ae8): the Context is generalised (SES is now just the closing illustration) and now leads with the encapsulation decision rather than the stateful/stateless detail.

docs/adr/0016-domain-action-custom-resource.md
laazyj: Done — set to **Proposed** in 986916b (pending a second instance to price it).

packages/ses/README.md
laazyj: Done in 986916b — each component (email identity, receipt rule set, receipt filters, actions) now opens with a 1–2 sentence description of what it is.

packages/ses/README.md
laazyj: Done in 986916b — `.addAction(key, action)` is documented as taking any CDK `IReceiptRuleAction`, with the `s3Action`/`lambdaAction`/… free functions framed as optional convenience helpers, plus a note to construct the CDK action directly for types we don't yet wrap.

packages/ses/README.md
laazyj: Done in 986916b — linked the "no CloudFormation resource" issue (aws-cdk#28823) and the SES *managing receipt rule sets* docs on the activation limitation.

packages/ses/package.json
laazyj: Done in 986916b — generalised to "Compose-native builders for AWS SES …"; the README notes receiving is the current cut with sending-side to follow.

packages/ses/src/actions.ts
laazyj: Done in 986916b — split into `src/actions/` (one file per action) with a barrel `index.ts`.

packages/ses/src/actions.ts
laazyj: Done in 986916b — documented as defaulting to CDK's default, `Event`.

packages/ses/src/actions.ts
laazyj: Done in 986916b — documented as defaulting to SES's default, UTF-8.

packages/ses/src/index.ts
laazyj: Done in 986916b — `warnIfNotReceivingRegion` is no longer exported (internal). The `SES_RECEIVING_REGIONS` and `RECEIVING_REGION_WARNING` constants remain, as they're useful for inspecting/suppressing the warning.

PR #307, opened 2026-07-16
build(lint): make eslint a cached per-project nx target

laazyj:
## Why

Lint had become the slowest gate: `npm run lint` was a single root `eslint .`
that re-lints the entire tree on every invocation with **no nx caching** —
unlike `build`, `test`, and `typecheck`, which are cached nx targets that
fast-succeed on unchanged projects. This makes lint a cached per-project nx
target while keeping coverage identical.

## Approach (and one it deliberately avoids)

nx derives every target in this repo from **package.json scripts** — there are
no inference plugins. So each package gets a `"lint": "eslint ."` script and nx
infers a cached `lint` target for it, exactly like `build`/`test`/`typecheck`.

I first tried the idiomatic `@nx/eslint` **inference plugin**, but it evaluates
the root flat config during *graph construction* (`ESLint.isPathIgnored`, to
skip projects with nothing to lint). That runs
`import "@composurecdk/eslint-plugin"` before the package is compiled, so **every
nx command fails on a fresh checkout** — it only appeared to work locally
because a stale `dist/` was lying around. Per-package scripts don't load the
config until lint actually runs, by which point `dependsOn` has built the
plugin. This is documented in `AGENTS.md` so it isn't re-introduced.

## What changed

- **Per-package `lint` scripts** — `"lint": "eslint ."` in each package.json;
  `npm run lint` / `lint:fix` run `nx run-many -t lint`.
- **`workspace-root` project** (root `project.json`) — a cached `lint` target
  for the loose top-level files no package owns (`eslint.config.mjs`,
  `scripts/**`, `vitest.config.base.ts`), so the whole repo is still linted.
- **`targetDefaults.lint`** — `dependsOn` the `@composurecdk/eslint-plugin`
  build (the flat config imports its compiled output) and keys its cache on
  that package's `src/**`, so a rule change busts dependent lint caches.
  (`dist` is gitignored → invisible to nx's file hasher, so source is the
  correct key.)
- **`nx.includedScripts: []` on the root package.json** — stops the root
  project from turning root scripts (`build`, `test`, `verify`, the root `lint`)
  into `workspace-root` targets, which would otherwise recurse.
- **Docs** — `AGENTS.md` and `docs/ci.md`.

## Results

| Run | Before (`eslint .`) | After (`nx run-many -t lint`) |
| --- | --- | --- |
| Warm (no changes) | ~18s, full re-lint | **<1s**, 25/25 from cache |
| One package changed | ~18s, full re-lint | only that package re-lints |
| Cold / CI | ~18s | comparable (24 projects in parallel) |

## Verification

Validated from a simulated fresh checkout (`rm -rf packages/eslint-plugin/dist
.nx`):

- `typecheck`, `build`, and `lint` all pass — the graph loads without evaluating
  the flat config (this is the exact failure the inference-plugin attempt hit).
- Cold lint runs all 24 projects (23 packages + `workspace-root`); warm run is
  25/25 from cache.
- A lint error in a package fails **only** that package's task; a root-file
  error fails `workspace-root:lint`.
- Editing a rule in `@composurecdk/eslint-plugin/src` rebuilds it and busts the
  dependent lint caches.
- `build` / `typecheck` / `test` run-many exclude `workspace-root`; it is
  outside the `packages/*` release glob.

## Notes / follow-ups

- No ADR: this reuses the existing package.json-script → cached-nx-target
  pattern rather than introducing a new mechanism.
- CI caching is unchanged. Runners are ephemeral, so cross-run cache reuse would
  need Nx Cloud or an `actions/cache` step on `.nx/cache` — a separate decision,
  out of scope here.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,10 +19,11 @@ Fix any issues before moving on. Use npm run lint:fix and npm run format to auto
 
 Use npx nx to run build/test scripts — this is an nx monorepo.
 
-Lint is an nx target too: `npm run lint` runs `nx run-many -t lint`, which caches per project so unchanged packages fast-succeed. Two things make this correct rather than merely fast:
+Lint is an nx target too: `npm run lint` runs `nx run-many -t lint`, which caches per project so unchanged packages fast-succeed. Each package carries a `"lint": "eslint ."` script, so nx infers a `lint` target the same way it infers `build`/`test`/`typecheck` from package.json scripts — add that line when you create a package. Three things make this correct rather than merely fast:
 
-- **Loose top-level files** (`eslint.config.mjs`, `scripts/**`, `vitest.config.base.ts`) belong to no package, so they are linted by the `workspace-root` project defined in the root [`project.json`](project.json). If you add a source file outside `packages/` and outside those globs, extend that project's `lint` target so it stays covered — the goal is that every file the old `eslint .` linted is still linted.
-- **The custom rules** in `@composurecdk/eslint-plugin` drive every package's lint result, so `targetDefaults.lint` in [`nx.json`](nx.json) both depends on that package's `build` (the config imports its compiled output) and lists its `src/**` as a lint input, so a rule change busts the dependent lint caches.
+- **Loose top-level files** (`eslint.config.mjs`, `scripts/**`, `vitest.config.base.ts`) belong to no package, so they are linted by the `workspace-root` project defined in the root [`project.json`](project.json). If you add a source file outside `packages/` and outside those globs, extend that project's `lint` target so it stays covered.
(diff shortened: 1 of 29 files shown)


AGENTS.md
```diff
@@ -19,6 +19,11 @@ Fix any issues before moving on. Use npm run lint:fix and npm run format to auto
 
 Use npx nx to run build/test scripts — this is an nx monorepo.
 
+Lint is an nx target too: `npm run lint` runs `nx run-many -t lint`, which caches per project so unchanged packages fast-succeed. Two things make this correct rather than merely fast:
+
+- **Loose top-level files** (`eslint.config.mjs`, `scripts/**`, `vitest.config.base.ts`) belong to no package, so they are linted by the `workspace-root` project defined in the root [`project.json`](project.json). If you add a source file outside `packages/` and outside those globs, extend that project's `lint` target so it stays covered — the goal is that every file the old `eslint .` linted is still linted.
```
laazyj: Drop the historical reference  ("old eslint"). Irrelevant.

laazyj (commented):
Minor change to documentation required.

PR #324, opened 2026-07-19
build(nx): run hot targets directly instead of via `npm run`

### LINKED ISSUE #323: CI: intermittent task failure — nx runs package scripts via `npm run`, and npm's config loader crashes under parallelism
## Symptom

CI's `Lint, Format, Typecheck, Test, Build` job intermittently fails on **one matrix leg** with nx reporting a task failed — e.g. `@composurecdk/eslint-plugin:typecheck` — while the **same commit passes on the other Node legs**. Re-running the job clears it.

Observed on PR #318, Node 22 (attempt 1 failed, attempt 2 green): `.../actions/runs/29676492406`.

## What actually failed

It is **not** a TypeScript error. nx's task output for the failing task was:

```
##[group]✅ nx run @composurecdk/eslint-plugin:build      ← build ran (tsc)
##[group]❌ nx run @composurecdk/eslint-plugin:typecheck
   Exit prior to config file resolving
   cause
   call config.load() before reading values
```

The `> tsc --noEmit` command header never prints and the task starts-and-fails in ~2 ms — **`tsc` never ran**. The error is thrown by **npm's own config module** (`@npmcli/config` — `call config.load() before reading values`), not by nx (the string is absent from nx's JS and native binaries) and not by the compiler.

## Root cause

nx runs package.json script targets through the package manager. Its `run-script` executor builds `npm run <script>` and spawns it (`node_modules/nx/dist/src/executors/run-script/run-script.impl.js`):

```js
const pm = getPackageManagerCommand();
let command = pm.run(options.script, …);   // → "npm run typecheck"
spawn(command, [], …);                      // the npm process that crashed
```

So every inferred script target (`build`, `typecheck`, `test`, `check:exports`, `lint`) spawns an `npm` process, and nx runs many in parallel. Under that concurrency an `npm` startup intermittently aborts inside its config loader (`Exit prior to config file resolving`), failing the task before the underlying tool runs. It is non-deterministic and unrelated to the diff under test.

Related upstream: the npm error is [npm/cli#8425](https://github.com/npm/cli/issues/8425) (closed **not planned**); reproduces on npm 11.x. Not fixed by upgrading nx (latest 23.1.0 still spawns `npm run` for inferred script targets), and not by upgrading npm (no patched version).

## Fix

Map each package's hot targets to `nx:run-commands` so nx runs the tool **directly** (`tsc --noEmit`, `eslint .`, `tshy`, …) instead of via `npm run <script>`, removing the npm subprocess entirely. The `package.json` `scripts` stay as the source of truth (and for `npm run` by humans); `targetDefaults` in `nx.json` still supply `dependsOn`/`cache`/`inputs`/`outputs`.

Note: `nx.json` `targetDefaults` **cannot** override an inferred target's executor (verified) — the override has to live per-package, in each `package.json`'s `nx.targets`.

PR to follow.

laazyj:
## What & why

Closes #323.

CI's typecheck/lint/build/test job intermittently fails on a single matrix leg with nx reporting a task failed (e.g. `@composurecdk/eslint-plugin:typecheck`) while the same commit passes on every other Node leg. It is **not** a compiler error — the underlying tool never runs. nx executes package.json script targets through the package manager (its `run-script` executor spawns `npm run <script>`), and under parallelism an `npm` startup intermittently crashes inside npm's own config loader:

```
Exit prior to config file resolving
cause
call config.load() before reading values     ← @npmcli/config, not nx, not tsc
```

(Upstream: [npm/cli#8425](https://github.com/npm/cli/issues/8425), closed *not planned*; not fixed by upgrading nx or npm.) See #323 for the full trace.

This maps each package's hot targets — `build`, `typecheck`, `test`, `check:exports`, `lint` — to `nx:run-commands` under the `package.json` `nx.targets` block, so nx runs the tool **directly** (`tsc --noEmit`, `eslint .`, `tshy`, …) with `cwd` pinned to the project root, eliminating the `npm` subprocess that crashes.

## Details

- The `package.json` `scripts` are unchanged — they stay the source of truth for each command and keep `npm run <script>` working for humans; `nx.targets` mirrors them.
- `nx.json` `targetDefaults` continue to supply each target's `dependsOn` / `cache` / `inputs` / `outputs` — the override only swaps the executor from the inferred `nx:run-script` to `nx:run-commands`. Verified that `targetDefaults` **cannot** override an inferred target's executor, which is why the override lives per package.
- Commands are copied verbatim from each package's existing script, so the per-package variants are preserved (`tshy` vs `tsc -p tsconfig.build.json` for `build`; `vitest run` vs `vitest run --passWithNoTests`; packages without a `build`/`check:exports` script get no such target).
- No `@nx/eslint` inference plugin is introduced — the `lint` override still runs `eslint .` at task time and keeps its `dependsOn` on `@composurecdk/eslint-plugin:build`, so the fresh-checkout ordering AGENTS.md describes is unchanged.
- AGENTS.md documents the pattern so new packages include the block.

## Validation — clean, CI-like state

Wiped `node_modules`, `npm ci`, `nx reset`, removed all `dist`/`coverage`, then ran the full gate. All green, and dual ESM/CJS `dist` artifacts are produced correctly:

| target | result |
| --- | --- |
| `build` (run-many) | ✅ 22 projects, `> tshy` / `tsc` run directly (no `npm run`) |
| `typecheck` | ✅ 23 projects, EXIT 0 |
| `lint` | ✅ 24 projects, EXIT 0 |
| `check:exports` | ✅ 20 projects, EXIT 0 |
| `test` | ✅ 23 projects, EXIT 0 |

## Checklist

- [x] Linked to an issue (#323)
- [x] `build`, `typecheck`, `test`, `check:exports`, `lint`, and `format:check` all pass locally from a clean `npm ci` + `nx reset`
- [x] No behaviour change to package sources — build-system config only

---
_Generated by [Claude Code](https://claude.ai/code/session_01QEDVnuJAM7zgUUgLNnxn9j)_


laazyj:
I'm not particularly happy with this and exploring whether we should be doubling-down on nx

PR #328, opened 2026-07-24
feat(lambda): harden dynamo stream event source failure handling

laazyj:
Closes #330.

## What & why

DynamoDB Streams support already largely existed in composureCDK (stream-enable API, `tableStreamArn`, the `dynamoEventSource` consumer, least-privilege `grantStreamRead`, and `IteratorAge`/failed-invoke/dropped-event alarms). The real gap was **failure handling**: CDK's `DynamoEventSource` defaults leave `retryAttempts`/`maxRecordAge` **unbounded** with **no dead-letter destination** and `bisectBatchOnError` off, so a single poison record can block a shard for the whole 24 h stream-retention window. This PR closes that gap to meet AWS Well-Architected guidance — without silently changing data-durability semantics — by extending the existing `lambda`/`dynamodb`/`examples` packages (streams is a native table prop + a Lambda consumer, so no new package).

**Lambda**
- **Shared stream defaults** — new `DEFAULT_STREAM_EVENT_SOURCE_PROPS` adds `bisectBatchOnError: true` (isolate a poison record) alongside start-at-tip, partial-batch reporting, and ESM metrics. Typed on `StreamEventSourceProps`, so a future `kinesisEventSource` reuses it verbatim.
- **Resolvable `onFailure` DLQ ergonomics** — `dynamoEventSource`'s `onFailure` is widened to `Resolvable<IQueue | IEventSourceDlq>`: pass a queue (or a `ref()` to a sibling `createQueueBuilder("dlq")` result) and it's wrapped in an `SqsDlq` and resolved alongside the table `ref` via `combine` (ADR-0015). `startingPosition` is now optional (the defaults supply it).
- **Stream failure-handling guard** — a suppressible synth-time relationship guard (`STREAM_DLQ_WARNING_ID`, ADR-0011) that warns only on the silent-data-loss configuration (retries/record-age bounded but no `onFailure` destination), filling the previously empty `dynamodb` guard slot. Written source-agnostically for reuse by a future stream kind.

**Docs & example**
- **ADR-0017** records the decision (bisect by default; DLQ by guard rather than silent finite retries; the Kinesis seam).
- New **`ComposureCDK-DynamoStreamProcessor`** example (table → stream → Lambda → SQS DLQ) with a synth test, plus lambda/dynamodb README updates.

**Kinesis synergy (out of scope, seam left ready):** the shared defaults constant and the guard are source-agnostic drop-ins for the future `@composurecdk/kinesis` package (#120). The one thing deliberately *not* done is hoisting the `onFailure`/`toDlq`/`combine` wiring into a shared resolution helper — premature against a single consumer; it's a clean extraction when Kinesis lands (noted in ADR-0017).

No new alarm constructs: `IteratorAge`/`FailedInvokeEventCount`/`DroppedEventCount` already cover the consumer, and the wired SQS DLQ brings its own depth alarm.

## Checklist

- [x] Linked to an issue — Closes #330
- [x] `npm run verify` passes locally (build, check:exports, lint, test)
- [x] Tests added/updated for the change (lambda event-source tests for the bisect default, DLQ wiring, the guard warn/silent/suppress cases; example synth test)

---

https://claude.ai/code/session_01C7gMQ99udZ3jgYzJRaS5Qo


docs/adr/0017-stream-event-source-failure-handling.md
```diff
@@ -0,0 +1,85 @@
...
+default `reportBatchItemFailures: true`; the rest was missing. The question is
+how much of this to bake into defaults versus leave to the user, given the
+project's rule that a default must be safe, individually overridable, and must
+not silently change data-durability semantics (docs/architecture.md, "Defaults").
+
+A second constraint: Kinesis stream support is planned. Whatever we add for
+DynamoDB should be the same code Kinesis reuses, since both sources share
+`StreamEventSourceProps` and the same failure model.
+
+## Decision
```
laazyj: This is a feature decision for how to support DynamoDB streams, it is not an architectural decision and does not require an ADR.

Drop ADR.

packages/examples/src/dynamo-stream-processor-app.ts
```diff
@@ -0,0 +1,98 @@
...
+ * - Bounded retries + an `onFailure` DLQ wired from a sibling `createQueueBuilder("dlq")`
+ * - Composing table + DLQ + processor + alert topic and routing all alarm
+ *   actions through `alarmActionsPolicy`
+ */
+export function createDynamoStreamProcessorApp(app = new App()) {
+  const stack = new Stack(app, "ComposureCDK-DynamoStreamProcessorStack");
+
+  const { alerts } = compose(
+    {
+      alerts: createTopicBuilder().displayName("Dynamo Stream Processor Alerts"),
```
laazyj: we don't need to demonstrate alerts in every example. Simplify it out.

packages/examples/bin/app.ts
```diff
@@ -21,6 +22,7 @@ createAgentVolumeApp(app);
 createCrudApiApp(app);
 createDnsZoneApp(app);
 createDualFunctionApp(app);
+createDynamoStreamProcessorApp(app);
```
laazyj, line 25: All examples must have a corresponding smoke test, otherwise they are not worth deploying.

laazyj (commented):
Some work required

docs/adr/0017-stream-event-source-failure-handling.md
laazyj: Agreed — dropped. `docs/adr/0017-*.md` is deleted, along with its index row and the lambda README link. The behaviour it described is documented in the lambda README (defaults table + the "Failure handling (dead-letter queue)" section) instead.

packages/examples/src/dynamo-stream-processor-app.ts
laazyj: Done — the SNS topic, `alarmActionsPolicy` call, and the alarm-routing assertion are gone. The example is now table → stream → Lambda → DLQ only.

packages/examples/bin/app.ts
laazyj: Added `packages/examples/test/smoke/dynamo-stream-processor.smoke.mjs`. It writes one item with a unique partition key and polls the consumer's log group for that marker, so it proves the whole path — stream delivery, the ESM wiring, and the execution role's `grantStreamRead` — not just that the stack is healthy.

One thing that needs your action before the next deploy-test run: the smoke write needs `dynamodb:PutItem`, which the deploy-test role did not have. I added a `DynamoDBPutItem` statement to `.github/cloudformation/github-oidc-role.yml`, ARN-scoped to the sandbox account like the existing `SQSSendMessage` one (DynamoDB has no tag-based authorization for data-plane actions). The `github-actions-oidc` stack needs redeploying to pick it up.

PR #327, opened 2026-07-24
feat(apigateway): add restApiGrants.invoke consumer-side grant helper

laazyj:
## What & why

Closes #326.

`@composurecdk/apigateway` shipped builders but, unlike `@composurecdk/dynamodb`'s `tableGrants`, no consumer-side grant helper. To let a principal invoke a REST API (e.g. a cross-account principal that SigV4-signs `execute-api` calls), consumers had to hand-roll the grant with `grantVia` + a statement builder + `arnForExecuteApi()`.

This adds a `restApiGrants` capability namespace mirroring the `tableGrants` shape:

```ts
createRoleBuilder()
  .assumedBy(principal)
  .grant(restApiGrants.invoke(ref("api", (r) => r.api)));
```

- **`restApiGrants.invoke(api, scope?)`** returns a `Grant<IGrantable>` via `grantVia` and adds `execute-api:Invoke` scoped to the API's `arnForExecuteApi()`. `IRestApi` is implemented by both `RestApi` and `SpecRestApi`, so one helper serves either builder's result — the same way `ITable` serves both DynamoDB builders.
- **Optional scoping** via `RestApiInvokeScope` (`{ method, path, stage }`, each defaulting to `*`) maps straight onto `arnForExecuteApi(method, path, stage)`. Default covers the whole API.

### On the implementation

aws-cdk-lib has no API-wide invoke grant to delegate to — `grantExecute` exists only on `Method`, not on `IRestApi`/`RestApiBase` — so the grant is assembled from the single well-known `execute-api:Invoke` action plus the construct's own `arnForExecuteApi()` for the resource ARN. This is a deliberate, narrow exception to ADR-0013's "defer to the construct's native `grant*` method" principle, documented in a new **ADR-0017** (bounded to a single canonical action + the construct-owned ARN).

### Changes

- `packages/apigateway/src/grants.ts` — `restApiGrants` + `RestApiInvokeScope`, exported from the package index
- `packages/apigateway/test/grants.test.ts` — default grant, scoped grant, and Ref-resolution (100% coverage)
- `docs/adr/0017-assemble-grants-when-no-native-method.md` + index entry
- `packages/apigateway/README.md` — "Invoke grants" section

Examples were reviewed; none of the existing API examples has an IAM-authorized API with a distinct caller principal, so the helper isn't retrofitted into any of them (that would change their teaching focus). No new example added per discussion.

## Checklist

- [x] Linked to an issue (Closes #326)
[...]

PR #346, opened 2026-07-27
feat(ci): comment on issues a release addressed

laazyj:
## What & why

An issue's watchers currently have no way to learn which version their fix shipped in. The release notes carry the links, but only in one direction — nothing points back at the issue.

`release-notify.yml` listens for a successful **Release** run and posts one comment per issue that release addressed:

> Addressed in **v0.9.1** — see the [release notes](https://github.com/laazyj/composureCDK/releases/tag/v0.9.1).
>
> Published to npm as `@composurecdk/*@0.9.1`.

### How issues are found

Two passes over the published release body, because neither alone is complete:

1. **`…/issues/N` links** — what nx renders from `Closes #N` footers in *commit messages*.
2. **`…/pull/N` links, expanded to each PR's linked closing issues** (GraphQL `closingIssuesReferences`). The PR template asks for `Closes #123` in the *PR body*, which never reaches the squash commit message, so nx never sees it.

Candidates are then filtered to real issues — `/issues/N` resolves a PR number just as happily, and the changelog does contain such links (`v0.9.0` renders PR #276 that way). Every comment carries a `<!-- composurecdk-release: vX.Y.Z -->` marker, so re-running posts nothing twice.

### Why a listener, not a job in `release.yml`

Same argument `coverage-comment.yml` makes for sitting outside `ci.yml`, plus two specific to this path: `release.yml`'s conclusion answers *"did the publish work?"*, so a rate-limited comment must not turn it red after the packages are live; and `release.yml` is the npm trusted-publisher anchor, the one file worth keeping free of unrelated write scopes. Listening also buys a `workflow_dispatch` re-run path for free — which is what the marker idempotency exists for.

## Verification

Dry-run against two real past releases (read-only; posts nothing):

```
$ node scripts/release-issue-comments.mjs --tag=v0.9.1 --dry-run
3 commented, 0 skipped, 0 failed.
  #279, #280, #326

$ node scripts/release-issue-comments.mjs --tag=v0.9.0 --dry-run
5 commented, 1 skipped, 0 failed.
  #269, #270, #278, #279, #281
  skipped: #276 — a pull request, not an issue
```

The PR-expansion path is confirmed separately (PR #327 → issue #326); on these two releases it resolves to issues the notes already linked directly, so the union is a no-op there.

`actionlint` clean; `npm run verify` passes.

⚠️ Like `coverage-comment.yml`, this workflow must be on the **default branch** to fire, so it is not exercised by this PR. The first live run is the next release — `Release Notify` can also be dispatched by hand against any tag if it needs a nudge.

## Checklist

- [ ] Tests added/updated for the change — n/a, `scripts/` has no test harness in this repo; verified by dry-run above

### DIFF
--- a/.github/workflows/release-notify.yml
+++ b/.github/workflows/release-notify.yml
@@ -54,14 +54,16 @@ jobs:
         id: release
         env:
           TAG: ${{ inputs.tag || github.event.workflow_run.head_branch }}
+        shell: bash
         run: |
-          case "$TAG" in
-            v[0-9]*.[0-9]*.[0-9]*) ;;
-            *)
-              echo "::error::Refusing to announce: '$TAG' is not a vX.Y.Z release tag."
-              exit 1
-              ;;
-          esac
+          # The same SemVer shape release-tag.yml parses out of the release
+          # commit subject — MAJOR.MINOR.PATCH with optional -prerelease and
+          # +build — so the two ends of the chain accept exactly one set of
+          # tags between them.
+          if [[ ! "$TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$ ]]; then
+            echo "::error::Refusing to announce: '$TAG' is not a vX.Y.Z release tag."
+            exit 1
+          fi
           echo "tag=$TAG" >> "$GITHUB_OUTPUT"
 
       - name: Checkout
--- a/.github/workflows/release-notify.yml
+++ b/.github/workflows/release-notify.yml
@@ -1,0 +1,77 @@
+name: Release Notify
+
+# Comments on every issue a release addressed, once the packages are live.
+#
+# A workflow_run listener rather than a job inside release.yml, for the same
+# reason coverage-comment.yml sits outside ci.yml — plus two specific to the
+# release path:
+#
+#   - release.yml's conclusion answers "did the publish work?". A rate-limited
+#     or otherwise failed comment must not turn it red after the packages are
+#     already on npm.
+#   - release.yml is the npm trusted-publisher anchor (trust is pinned to its
+#     job_workflow_ref), so it is the one file worth keeping free of unrelated
+#     write scopes.
+#
+# Listening also buys a re-run path for free — see the workflow_dispatch input
+# below, which is what the script's marker idempotency is built for.
+
+on:
+  workflow_run:
+    workflows: [Release]
+    types: [completed]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag to announce (e.g. v0.9.1)."
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: release-notify-${{ inputs.tag || github.event.workflow_run.head_branch }}
+  cancel-in-progress: false
+
+jobs:
+  notify:
+    name: Comment on released issues
+    # Never announce a release that failed to publish. A manual dispatch names
+    # the tag itself, so it is trusted to know what it is doing.
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read # gh release view
+      issues: write # post the comments
+      pull-requests: read # resolve each PR's linked closing issues
+    steps:
+      # release.yml only ever runs on a tag push, which sets head_branch to the
+      # tag name — validated here rather than trusted, the same way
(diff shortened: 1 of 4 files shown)


.github/workflows/release-notify.yml
```diff
@@ -0,0 +1,77 @@
...
+      # release.yml only ever runs on a tag push, which sets head_branch to the
+      # tag name — validated here rather than trusted, the same way
+      # coverage-comment.yml validates the PR number it reads from an artifact.
+      - name: Resolve release tag
+        id: release
+        env:
+          TAG: ${{ inputs.tag || github.event.workflow_run.head_branch }}
+        run: |
+          case "$TAG" in
+            v[0-9]*.[0-9]*.[0-9]*) ;;
```
laazyj: Wouldn't this catch strange tags like "v.." or "var"?

Wouldn't `^v[0-9]+\.[0-9]+\.[0-9]+$` be a more accurate regex?

scripts/release-issue-comments.mjs
```diff
@@ -0,0 +1,233 @@
...
+ *      template asks for `Closes #123` in the *PR body*, which never reaches
+ *      the squash commit message, so nx cannot see those at all.
+ *
+ * Candidates are then filtered to real issues: GitHub's `/issues/N` URL
+ * resolves a PR number just as happily, and the changelog does contain such
+ * links.
+ *
+ * Shells out to the `gh` CLI rather than taking an Octokit dependency — same
+ * trade as `smoke-test.mjs` makes with the AWS CLI. Requires `GH_TOKEN` (or an
+ * authenticated `gh`) with `issues: write`.
```
laazyj: I don't understand the relevance of "Octokit". Explain in a reply then consider if this is really a useful (rather than distracting) comment to have in the file.

laazyj (commented):
Merge conflicts and 2 comments to consider

.github/workflows/release-notify.yml
laazyj: Good instinct, wrong examples — and the regex you proposed is too strict. Both fixed anyway.

**Neither of your cases matches.** The pattern is `v[0-9]*.[0-9]*.[0-9]*`, and `[0-9]` immediately after the `v` requires a digit there. `v..` has `.` in that position and `var` has `a`, so both fall through to the `*)` arm and exit 1.

**But you're right that it's loose.** `*` in a glob spans any character, including `/` and spaces, so these all pass:

```
v..                    no
var                    no
v1x.2y.3z              MATCH   ← junk
v1.2.3/../x            MATCH   ← path traversal
v1.2.3 --repo evil/x   MATCH   ← argument injection into `gh release view`
```

That last one is the one that actually matters: `$TAG` is passed to `gh release view`.

**On `^v[0-9]+\.[0-9]+\.[0-9]+$`** — it closes all three, but it also rejects `v0.9.2-rc.1` and `v1.2.3+build`, which the rest of the chain accepts: `release-tag.yml`'s parse allows prerelease and build identifiers, and `release.yml` triggers on `v*.*.*`. A tag valid enough to publish from would then be refused an announcement.

So I've used the SemVer regex `release-tag.yml` already parses out of the release commit subject, which makes both ends of the chain agree on exactly one set of tags:

```bash
if [[ ! "$TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$ ]]; then
```

Verified against the table above: everything in the junk column is rejected, `v1.2.3`, `v0.9.2-rc.1` and `v1.2.3+build` accepted. `shell: bash` is now explicit on the step, since `=~` is not POSIX.

scripts/release-issue-comments.mjs
laazyj: Octokit is GitHub's official JavaScript SDK (`@octokit/rest`) — the thing you'd normally reach for to call the GitHub API from Node, and the obvious alternative to shelling out to `gh`. I named it to pre-empt "why is this driving a CLI instead of using a proper client?".

You're right that it's distracting. Nothing in this repo uses Octokit or ever has, so the comment introduces a library that appears nowhere else in the codebase purely to say we didn't use it — it explains the road not taken rather than the constraint that made the decision. The constraint is that `scripts/` is dependency-free, which is a rule a reader can check and apply to their own change; "we rejected Octokit" is trivia.

Rewritten to state it positively, and to point at the precedent rather than the rejected option:

```
 * Drives the `gh` CLI, so `scripts/` stays dependency-free — the same way
 * `smoke-test.mjs` uses the AWS CLI. Requires `GH_TOKEN` (or an authenticated
 * `gh`) with `issues: write`.
```

`smoke-test.mjs` carries the same rationale for the AWS CLI, so anyone touching either one lands on the same rule.

Issue #18, opened 2026-04-13
Consider default DLQ in SNS


laazyj:
Once an SQS or other suitable package is available

PR #349, opened 2026-07-27
docs(sns): demonstrate subscription dead-letter queues in the order processor

laazyj:
Closes the question in #18 (default DLQ in SNS) by documenting and
demonstrating the robust shape rather than defaulting it. Auto-creating a
DLQ would materialise a second billable queue — with its own retention,
access policy, alarms and deletion behaviour — from a topic subscription,
which is not what defaults do here: they set properties on resources the
caller asked for.

Extends ComposureCDK-OrderProcessorStack into an SNS -> SQS -> Lambda
system: an `orderEvents` intake topic fans out to the existing `orders`
work queue via `TopicBuilder.addSubscription`, with an explicit
`createQueueBuilder("dlq")` sibling attached to the subscription through
`combine`. The DLQ picks up the dlq role's 14-day retention and depth
alarm, and the topic's redrive alarms now have data to report.

- Synth tests cover the subscription, its redrive policy, the SNS
  send-message grant CDK adds to the DLQ, and the new alarm count.
- The smoke test now publishes through the topic as well as sending
  directly to the queue, proving the fan-out path end to end.
- SNS README expands "Subscription reliability" with the rationale and a
  compose snippet.

The deploy-test OIDC role gains `sns:Publish` on `ComposureCDK-*` topics
for the new smoke check. `.github/cloudformation/github-oidc-role.yml` is
deployed by hand — it must be redeployed before the next deploy-test run
or the check fails with AccessDenied.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01MDiCumXKUAubBMGvUdo5Le


.github/cloudformation/github-oidc-role.yml
```diff
@@ -319,6 +319,19 @@ Resources:
             Resource:
               - !Sub arn:${AWS::Partition}:sqs:*:${AWS::AccountId}:*
 
+          # SNS — smoke test publishes an order event to the order-processor
+          # intake topic to verify the SNS -> SQS -> Lambda delivery path.
+          # The example leaves the topic unnamed, so CloudFormation generates
```
laazyj: Generated name is too fragile, don't rely on it. Remove the comment and expand the resources that can be published to. Ideally limit via Tags but, if not possible, then allow publish to any topic in the account (similar pattern exists for SQS).

Remember to update the related comment in `ci.md`

packages/examples/src/order-processor-app.ts
```diff
@@ -26,7 +33,12 @@ import { createQueueBuilder, type QueueBuilderResult } from "@composurecdk/sqs";
  * - Adding a workload-specific alarm via `addAlarm`
  * - Wiring the queue to a `createFunctionBuilder` consumer via
  *   `sqsEventSource` and a `ref` to the sibling queue
- * - Composing the queue alongside `createTopicBuilder` and routing all
+ * - SNS → SQS fan-out via `TopicBuilder.addSubscription`, with a
+ *   **caller-owned dead-letter queue** on the subscription (see
+ *   "Subscription reliability" in the SNS README): the DLQ is an explicit
+ *   `createQueueBuilder("dlq")` sibling, not something the topic builder
+ *   conjures, and `combine` assembles the subscription from both queues
```
laazyj: "Not something the topic builder conjures" is irrelevant in this situation. The decision is made not to auto-generate DLQs, this comment does not justify the decision (too much info for this site) and re-opens a door to re-litigate the question.

Remove it.

packages/sns/README.md
```diff
@@ -215,9 +215,43 @@ const system = compose(
...
+          orders: ref<QueueBuilderResult>("orders"),
+          dlq: ref<QueueBuilderResult>("ordersDlq"),
+        },
+        ({ orders, dlq }) => new SqsSubscription(orders.queue, { deadLetterQueue: dlq.queue }),
+      ),
+    ),
+  },
+  { orders: [], ordersDlq: [], orderEvents: ["orders", "ordersDlq"] },
+);
+```
```
laazyj, line 238: Shorten this example to just the bits that need demonstrating, it is too long for documentation.

laazyj (commented):
Some documentation improvements to be made

.github/cloudformation/github-oidc-role.yml
laazyj: Done — dropped the generated-name comment and widened the statement to `arn:${AWS::Partition}:sns:*:${AWS::AccountId}:*`, mirroring the SQS one (same reasoning: CloudFormation's `aws:cloudformation:*` system tags don't propagate to SNS topics in a form IAM can evaluate via `aws:ResourceTag`, and the OIDC trust policy already confines the role to the `sandbox` environment).

`docs/ci.md` updated too — the security note now groups SQS, SNS, and DynamoDB as ARN-scoped to the sandbox account, and the claim about the stack-name prefix is gone.

packages/examples/src/order-processor-app.ts
laazyj: Removed. The bullet now just states what the example does — an explicit `createQueueBuilder("dlq")` sibling assembled through `combine` — and points at the SNS README for the reasoning.

I also trimmed the same re-litigating clause from the inline comment on `.addSubscription` further down the file ("a DLQ is a real, billable, caller-owned resource with its own retention and alarms"), leaving what SNS actually does with the queue and where it is declared.

packages/sns/README.md
laazyj: Cut down. The snippet lost its import block and the `queueName` calls, and the `combine` refs are on one line — it is now just the three entries and the dependency map, which is the part that needs demonstrating.

The prose above it shrank with it: the paragraph justifying the boundary (billing, encryption, deletion behaviour, the cross-reference to `docs/architecture.md#defaults`) is down to one sentence — a DLQ needs an owner, so it is declared as a sibling and gets the dead-letter defaults and depth alarm.

Issue #336, opened 2026-07-26
Non-ASCII in Description fields is silently transliterated to '?' by CloudFormation, causing perpetual no-op diffs


laazyj:
## Summary

Non-ASCII characters (em-dash `—`, en-dash, curly quotes, …) in `Description` fields are accepted at synth, then **silently transliterated to `?` by CloudFormation at deploy**. The deployed template therefore never matches the synthesised one, and `cdk diff` reports a change on *every* run — forever — even when nothing has actually changed.

This is the same root cause as the `GroupDescription` failure that motivated [ADR-0010](docs/adr/0010-aws-property-constraints.md), but with a worse failure mode: EC2 rejected the value outright with `CREATE_FAILED: Character sets beyond ASCII are not supported`, whereas the template-level `Description` is accepted and quietly rewritten. There is no error to notice — just a permanent no-op diff.

## Evidence

From a downstream consumer's PR: https://github.com/laazyj/jasonduffett.net/pull/96#issuecomment-5083048310

```shell
Template
[~] Description Description: jasonduffett.net ? static site on CloudFront + S3. to jasonduffett.net — static site on CloudFront + S3.
```

Left side (`?`) is what CloudFormation stored; right side (`—`) is what `cdk synth` produces. The stack is reported as `0 to add, 2 to update, 0 to destroy` on a PR that changed nothing about the stack description. The diff is unresolvable by the consumer without discovering the transliteration and hand-stripping the character.

## Coverage gap

The ADR-0010 catalogue exists but only covers a few properties. Fields carrying free-text that composureCDK emits, with no constraint applied:

| Field | Path | Status |
| --- | --- | --- |
| Stack `Description` | `packages/cloudformation/src/stack-builder.ts` (`StackProps.description` passthrough) | no validation |
| CloudWatch `AlarmDescription` | `packages/cloudwatch/src/alarm-definition-builder.ts` → `create-alarms.ts:53` | no validation (only `AlarmName` is checked, `packages/cloudwatch/src/alarm-name.ts`) |
| `CfnOutput` `Description` | `packages/cloudformation/src/outputs.ts:111` | no validation |

`StackBuilder` is a plain `StackProps` passthrough, so `description` is never seen by a validator at all.

## The library ships the bug in its own defaults

Four built-in alarm descriptions contain em-dashes, so any consumer using these alarm defaults inherits a perpetual diff on those alarms without writing a single non-ASCII character themselves:

[...]

PR #351, opened 2026-07-31
fix: strip non-ASCII from descriptions emitted into templates

laazyj:
## What & why

CloudFormation stores template text as ASCII and silently transliterates anything outside that set to `?` at deploy time. The deployed template then never matches the synthesised one, so `cdk diff` reports a change on **every** run, forever, on a stack nobody touched. Unlike the `GroupDescription` failure that motivated [ADR-0010](docs/adr/0010-aws-property-constraints.md) — where EC2 rejected the value outright — there is no error to notice.

Six library-shipped alarm descriptions carried an em-dash, so a consumer using those defaults inherited a perpetual diff without writing a single non-ASCII character themselves:

| Package | Alarm |
| --- | --- |
| `cloudfront` | CloudFront function throttles |
| `ec2` | Instance CPU credit balance |
| `ec2` | EBS volume burst balance |
| `ec2` | VPC interface endpoint packets dropped |
| `sqs` | DLQ oldest-message age |
| `sqs` | DLQ messages present |

The two SQS ones are not in the issue's list but are the same defect, found by scanning every package for non-ASCII outside comments.

Four example stacks carried the same character in resource descriptions (`crud-api-app`, `dual-function-app`, `dynamo-stream-processor-app`, `order-processor-app`). CI deploys those, so they had the same perpetual diff against live AWS.

Only template-bound strings are changed. Em-dashes in JSDoc and in synth-time error messages are untouched — neither reaches a template.

### Regression guard

`packages/examples/test/ascii-templates.test.ts` synthesises every example stack and asserts the emitted template JSON is printable ASCII. To stop that guard silently degrading as examples are added, the app registry moved out of `bin/app.ts` into `packages/examples/src/apps.ts`; the entrypoint and the test now share one list, so a new example is covered by construction. `AGENTS.md` step 2 points at the new file.

This guard only reaches library defaults the example stacks actually instantiate — `acm`, `budgets`, `route53` health checks and the standard (non-DLQ) SQS alarm profile are ASCII-clean today but not covered by it. Closing that gap properly is the job of the follow-up, not a second partial mechanism here.

### Scope

This is the immediate fix for the shipped defaults. It deliberately does **not** add throwing validation at the builders: that would turn working (if diff-noisy) deployments into synth failures for existing consumers. General protection for invalid CloudFormation values — an opt-in Aspect-based policy, per ADR-0010's named follow-up and [ADR-0002](docs/adr/0002-policies.md)'s Policy shape — is planned separately.

Refs #336 (does not close it — the underlying coverage gap remains open).

## Checklist

- [ ] If it adds an example stack: registered, listed in the examples README, and covered by a smoke test that exercises its runtime behaviour — n/a, no new example

---
_Generated by [Claude Code](https://claude.ai/code/session_01LoEijXK47YPhca1Zbp2C3P)_


packages/examples/bin/app.ts
```diff
@@ -1,35 +1,4 @@
...
-createMockApiApp(app);
-createMultiStackApp(app);
-createNeptuneGraphApp(app);
-createOpenApiPetstoreApp(app);
-createOrderProcessorApp(app);
-createStaticWebsiteApp(app);
-createTaggedSystemApp(app);
-
-app.synth();
+buildExampleApp().synth();
```
laazyj, line 4: I'm curious about this refactor (splitting two app.ts). I think I understand the justification but can you explain it and whether it is a common pattern?

No code change, just reply.

laazyj (commented):
No changes, but an open question to respond to

packages/examples/bin/app.ts
laazyj: Yes — it's the standard "bin is a thin wrapper over an importable function" split, and here it's forced rather than stylistic.

**Why it's forced.** `bin/app.ts` ends in a top-level `app.synth()`. That makes it a *script*, not a module: importing it from a test would run a full synth to `cdk.out` as an import side effect, before a single assertion. So the test can't reuse the entrypoint's list — it has to redeclare it. That's exactly what the first draft of this PR did, and it's the wrong shape: `AGENTS.md` step 2 tells the next contributor to register a new example in `bin/app.ts` and says nothing about the test, so the very next example added would be deployed by CI and *silently* excluded from the ASCII guard. The test would still pass — over 13 of 14 stacks. A guard that quietly narrows as the thing it guards grows isn't a guard.

Moving the registry into `src/apps.ts` makes it a plain function with no side effects. `bin/app.ts` becomes `buildExampleApp().synth()` — the one line that *is* the entrypoint's job — and the test calls `buildExampleApp()` and synths into its own outdir. One registration site, and a new example is covered by construction rather than by remembering.

**Is it common?** Very, in two forms:

- **Node CLIs generally:** the `bin/` file parses argv and calls a library function; everything testable lives in `src/`. Keeping the executable side effect out of the module graph is the whole point.
- **CDK specifically:** the AWS docs' own guidance is that `bin/*.ts` should do little more than `new App()`, instantiate stacks, and `synth()`. Multi-stack repos routinely factor the "which stacks exist" list into an importable module so tests and the entrypoint agree. `cdk.json`'s `app` command still points at `bin/app.ts`, so nothing changes about how CI deploys.

The one thing I did add beyond the mechanical move was updating `AGENTS.md` step 2 to point at the new file, so the checklist matches reality again — otherwise the split just relocates the drift.

---

PR #356, opened 2026-08-04
feat(apigateway): accept a resolvable apiDefinition on the spec REST API builder

laazyj:
## What & why

Closes #355.

A spec-driven REST API whose integrations name sibling resources — the ARN of the Lambda to invoke, the role API Gateway assumes to invoke it — could not be built inside `compose`. `apiDefinition` accepted only a concrete `ApiDefinition`, and `SpecRestApiBuilder.build()` never declared the `context` parameter, so the values the specification needs did not exist when the builder was configured. The reporter's workaround was to build the `SpecRestApi` outside `compose` and lose the dependency graph.

**The change:** widen `apiDefinition` to `Resolvable<ApiDefinition>` and resolve it against the build context — the three-step pattern `docs/architecture.md` prescribes and that `DistributionBuilder.certificate` and `CertificateBuilder.validationZone` already follow (`Omit<CdkProps, "x">` + a `Resolvable` re-declaration, resolved in `build`).

```ts
api: createSpecRestApiBuilder()
  .restApiName("PetStore")
  .apiDefinition(
    combine(
      {
        handler: ref<FunctionBuilderResult>("handler"),
        gatewayRole: ref<RoleBuilderResult>("gatewayRole"),
      },
      ({ handler, gatewayRole }) =>
        ApiDefinition.fromInline(
          withIntegration(spec, {
            functionArn: handler.function.functionArn,
            credentialsArn: gatewayRole.role.roleArn,
          }),
        ),
    ),
  ),
```

The substitution stays a plain function the consumer can test on its own, and a definition assembled from several siblings needs no further builder support — `combine` returns an ordinary `Ref` and drops into the widened seam (ADR-0015). No new seam, no spec-parsing in the builder, and `ApiDefinition.fromAsset`/`fromBucket` keep working unchanged.

### Notes on the details

- **Rejected the issue's proposed `transformSpec(spec, context)` callback.** It would be a second resolution mechanism with its own context object, duplicating `ref`/`combine`, and it assumes the definition is a parseable object — untrue for asset- and bucket-backed specs. Discussed on the issue.
- **The definition resolves before any construct is created**, so an unresolvable ref fails without leaving an orphan access log group grafted onto the scope.
- **A missing definition now throws** `SpecRestApiBuilder "<id>" requires an apiDefinition. Call .apiDefinition() with an ApiDefinition or a Ref to one.` — the repo's standard wording — instead of failing inside CDK. `apiDefinition` is declared optional on the props type to match every other widened-and-checked prop in the repo; the builder enforces it at build.
- **No ADR.** Applying an existing pattern to one more prop is not architecturally significant (AGENTS.md); documented in the package README instead.
- **Semver:** the setter widens (compatible). The getter `.apiDefinition()` now returns `Resolvable<ApiDefinition>`, a minor break for anyone reading it back.

### Docs

- `packages/apigateway/README.md` — new **Spec REST API Builder** section (the package had none), including the resolvable-definition pattern, the `Aws.*`-vs-`Stack.of(scope)` gotcha, and the inline-body template size caveat.
- `docs/architecture.md` §Resolvable — one paragraph noting that a `.map()` transform sees the build context but not the scope, so `Aws.*` pseudo-parameters are how to reach account/region/partition from one. That is a property of the `Ref` mechanism, not of this builder, so it belongs in the shared doc.

### Tests

Six cases in `spec-rest-api-builder.test.ts`: a `ref` resolved against the context, a `combine` of two siblings, a CDK token surviving into the synthesised `Body` as an `Fn::Join`, resolution driven by `compose` rather than a hand-passed context, an undeclared-dependency ref, and the missing-definition error. Package coverage stays at 100%.

## Checklist

- [x] If it adds an example stack: registered, listed in the examples README, and covered by a smoke test that exercises its runtime behaviour — n/a, no example stack here; a Lambda-backed example follows in a separate PR

---
_Generated by [Claude Code](https://claude.ai/code/session_01L3a7Z1pDEpLEL2H8z7vzJ3)_

### DIFF
--- a/packages/apigateway/README.md
+++ b/packages/apigateway/README.md
@@ -33,9 +33,6 @@ Every [RestApiProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws
 `createSpecRestApiBuilder` builds a [SpecRestApi](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.SpecRestApi.html) — a REST API whose resources, methods and integrations come entirely from an OpenAPI specification rather than from `addResource`/`addMethod` calls. Every [SpecRestApiProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.SpecRestApiProps.html) property is a fluent setter, and the same secure defaults, access logging and recommended alarms apply as for `createRestApiBuilder`.
 
 ```ts
-import { ApiDefinition } from "aws-cdk-lib/aws-apigateway";
-import { createSpecRestApiBuilder } from "@composurecdk/apigateway";
-
 const api = createSpecRestApiBuilder()
   .restApiName("PetStore")
   .apiDefinition(ApiDefinition.fromInline(petstoreSpec))
@@ -49,27 +46,23 @@ A model-first spec — a Smithy or OpenAPI export, or a hand-written document �
 `apiDefinition` therefore accepts a `Resolvable<ApiDefinition>` — a concrete definition, or a `ref`/`combine` that produces one at build time. The specification stays declarative, the substitution stays a plain function you can test on its own, and the whole API stays inside `compose`:
 
 ```ts
-import { ApiDefinition } from "aws-cdk-lib/aws-apigateway";
-import { combine, compose, ref } from "@composurecdk/core";
-import { createSpecRestApiBuilder } from "@composurecdk/apigateway";
-import {
-  createFunctionBuilder,
-  functionGrants,
-  type FunctionBuilderResult,
-} from "@composurecdk/lambda";
-import { createServiceRoleBuilder, type RoleBuilderResult } from "@composurecdk/iam";
+// Substitution is an ordinary function of its inputs — no CDK scope, no builder.
+const withIntegration = (spec: object, arns: Record<string, string>) =>
+  JSON.parse(
+    Object.entries(arns).reduce(
+      (doc, [name, arn]) => doc.split(name).join(arn),
+      JSON.stringify(spec),
+    ),
+  );
 
 compose(
   {
-    handler: createFunctionBuilder()
-      .runtime(Runtime.NODEJS_22_X)
-      .handler("index.handler")
-      .code(code),
+    handler: createFunctionBuilder().handler("index.handler").code(code),
 
     // The role API Gateway assumes to invoke the handler — an ordinary
     // sibling, wired with a consumer-side grant (ADR-0013).
     gatewayRole: createServiceRoleBuilder("apigateway.amazonaws.com").grant(
-      functionGrants.invoke(ref<FunctionBuilderResult>("handler", (r) => r.function)),
+      functionGrants.invoke(ref("handler", (r: FunctionBuilderResult) => r.function)),
     ),
 
     api: createSpecRestApiBuilder()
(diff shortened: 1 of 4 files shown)


packages/apigateway/README.md
```diff
@@ -28,6 +28,79 @@ Every [RestApiProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws
...
+```ts
+import { ApiDefinition } from "aws-cdk-lib/aws-apigateway";
+import { combine, compose, ref } from "@composurecdk/core";
+import { createSpecRestApiBuilder } from "@composurecdk/apigateway";
+import {
+  createFunctionBuilder,
+  functionGrants,
+  type FunctionBuilderResult,
+} from "@composurecdk/lambda";
+import { createServiceRoleBuilder, type RoleBuilderResult } from "@composurecdk/iam";
```
laazyj: Shorten the documentation by omitting imports, they don't add much to comprehension but take a lot of vertical space

packages/apigateway/README.md
```diff
@@ -28,6 +28,79 @@ Every [RestApiProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws
...
+      .restApiName("PetStore")
+      .apiDefinition(
+        combine(
+          {
+            handler: ref<FunctionBuilderResult>("handler"),
+            gatewayRole: ref<RoleBuilderResult>("gatewayRole"),
+          },
+          ({ handler, gatewayRole }) =>
+            ApiDefinition.fromInline(
+              withIntegration(petstoreSpec, {
```
laazyj, line 78: withIntegration is undefined and hides important details. Either inline a super simple version or provide an implementation so the example is comprehensible.

Be conscious of the size of the doc and consider parts of the snippet that could be dropped without reducing the pattern being communicated.

laazyj (commented):
Some small doc improvements before shipping please

packages/apigateway/README.md
laazyj: Done in b44f1e3 — imports dropped from both snippets in this section. The short `createSpecRestApiBuilder` example above lost its two-line import block too, since the same reasoning applies.

I left the imports in the "Invoke grants" and other pre-existing sections alone rather than widening this PR into a README-wide sweep. Happy to do that as a follow-up if you want the whole file consistent.

---

packages/apigateway/README.md
laazyj: Fair — `withIntegration` was doing the most interesting work in the snippet and I'd hidden it. It's now inlined as the four lines it actually is:

```ts
// Substitution is an ordinary function of its inputs — no CDK scope, no builder.
const withIntegration = (spec: object, arns: Record<string, string>) =>
  JSON.parse(
    Object.entries(arns).reduce((doc, [name, arn]) => doc.split(name).join(arn), JSON.stringify(spec)),
  );
```

Seeing it written out also makes the placeholder map visible at the call site (`"${PetFunction.Arn}": handler.function.functionArn`), which was previously buried behind `{ functionArn, credentialsArn }`.

On size: the section is net shorter despite gaining the implementation — imports gone, and the handler trimmed to `createFunctionBuilder().handler("index.handler").code(code)` since its runtime and memory settings teach nothing here. What I kept and why:

- **`gatewayRole`** — it is the second sibling, and the reason this needs `combine` rather than a plain `ref`. Dropping it would leave the snippet unable to motivate its own shape.
- **The dependency map** — the point of the pattern is that the wiring is declared data.

I also aligned the example stack in #357 to this same `Record<string, string>` shape so the README and the code it links to now read identically, and rebased #358 on top.

---

PR #366, opened 2026-08-06
feat(lambda): invoke a function during deployment with `.invokeOnDeploy()`

### LINKED ISSUE #365: feat(lambda): compose-native deploy-time invocation (CDK Trigger) — run a Function during deployment and fail the deployment if it fails
## Problem / use case

A customer wants to invoke a Lambda function **as part of a CloudFormation deployment** and have the deployment fail if that function fails.

Their concrete case: during deployment, call an external request/response HTTP API — a synchronous call whose result decides whether the release is valid. Shapes of the same need we should expect to cover:

- **Register/announce a release** with an external service (release tracker, feature-flag service, CMDB) and abort the deploy if the registration is rejected.
- **Post-deploy verification**: call the API the stack just deployed and fail the stack if it doesn't answer correctly, so a broken release rolls back rather than sitting live.
- **Seeding**: write reference data through a service API once the backing resources exist, and fail if seeding fails — a half-seeded environment should never be reported as a successful deploy.

The common thread is that the call is **synchronous and consequential**: its outcome must gate the deployment, and a failure must produce a rolled-back stack and a diagnosable message, not a green deploy.

There is nothing in the library for this today. `@composurecdk/custom-resources` covers *SDK calls that have no CFN resource*, which is a different job — it invokes an AWS API from a provider Lambda, not the consumer's own function. Nothing in `@composurecdk/lambda` invokes a function at deploy time.

## Requirements

1. **Deploy-time invocation of a function built by `createFunctionBuilder`.** The handler is an ordinary component of the composed system, wired by `ref(...)` like any other cross-component reference — not a construct the consumer has to reach for outside `compose()`.
2. **Synchronous by default.** The deployment waits for the function's response. Fire-and-forget must not be the default; if it is offered at all, choosing it must be explicit at the call site.
3. **A failing function fails the deployment.** If the handler throws, times out, or is not invocable, the deployment must fail and roll back. This is the requirement that rules out the obvious workaround (see Alternatives) and should be the primary thing the tests assert.
4. **The failure is diagnosable.** The handler's error message must reach the CloudFormation stack events; a reviewer reading the failed deploy should see why the API call failed without opening CloudWatch.
5. **Ordering is expressed as data in the compose graph.** Consumers must be able to say "run this after these components exist" using refs, consistent with how the rest of the library wires relationships — not procedural `node.addDependency` glue in an `afterBuild` hook. Requirement (1) covers the handler dependency; this covers everything else the call needs in place first (the API, the table, the permissions).
6. **A bounded, configurable wait.** The maximum time the deployment will wait for the handler is configurable, and the relationship to the handler's own timeout is enforced or at least surfaced — the wait must not be shorter than the handler's timeout, or a slow API is reported as the wrong failure.
7. **Re-invocation semantics are explicit and documented.** Consumers must be able to tell whether a given deploy will invoke the function again. Both behaviours have real use cases — "only when the handler changes" for idempotent registration, "every deploy" for verification — and the default must be stated in the README rather than inherited silently.
8. **Delete-time behaviour is documented.** Whether the function runs on stack deletion must be stated, and if it does not, the README should point at the existing custom-resource path for teardown-time work.
9. **IAM for the invocation is handled by the library**, scoped to the target function. Consumers should not have to hand-write an invoke policy; the handler's own permissions remain the function builder's business (`.configureRole`).
10. **The correctness trap is documented.** A handler that *returns* an error object rather than throwing is a successful invocation and will not fail the deployment. Whatever we ship must call this out prominently — it is the most likely way a consumer gets a silently-green deploy — ideally alongside a worked example in the package README.

## Non-goals

- Scheduled or event-driven invocation — that is `@composurecdk/events` and the event-source seam.
- Replacing `@composurecdk/custom-resources`. That remains the right tool for SDK calls with no CFN resource, and for anything needing create/update/**delete** semantics.
- Long-running orchestration during deploy (waiters, polling for eventual consistency). If the call cannot complete inside a bounded synchronous window, this is not the mechanism.

## Alternatives considered

**Raw `aws-cdk-lib/triggers.Trigger` in an `afterBuild` hook or a hand-written `Lifecycle` adapter.** This works today — `Trigger`'s provider checks the invoke response's status code and `FunctionError` and throws, which is exactly requirement (3). But every consumer re-derives the same ~15 lines of adapter, has to know that `InvocationType.REQUEST_RESPONSE` is what makes the failure propagate, and has to discover the re-invocation and delete-time semantics from the CDK source. That is the boilerplate this issue is asking to absorb.

**`createAwsCustomResourceBuilder()` with a `Lambda.invoke` call.** Fails requirement (3) and does so silently, which is the strongest argument for solving this properly. The `AwsCustomResource` provider only reports `FAILED` when the *SDK call* throws; a handler that throws still returns HTTP 200 with a `FunctionError` field, so the deployment goes green while the API call failed. A consumer reaching for the nearest existing tool in the library lands here and gets the wrong answer.

**Do nothing; document the raw pattern in the `@composurecdk/lambda` README.** Cheapest option, and defensible if this is a one-customer need. It leaves requirements (5) and (9) to each consumer and keeps the `AwsCustomResource` trap one wrong turn away, but it may be the right first step if we want a second use case before committing to API surface.

## Open questions for maintainer judgement

- **Placement.** `@composurecdk/lambda` (it is a thing you do to a function), or a separate package (it is a deployment-lifecycle concern and would pull `aws-cdk-lib/triggers` into lambda's surface)? Either way it needs the new-package treatment if it is separate — dual ESM/CJS, `DUAL_PACKAGES` registration, a `cdk-floors.json` entry.
- **Shape.** A builder (`create...Builder()`), or a function returning a `Lifecycle` component? It creates one construct and has no meaningful props surface of its own, which is a weaker fit for the builder pattern than most of the library.
- **Does this clear the examples bar?** Per AGENTS.md, an example must demonstrate a system rather than a resource. "Deploy an API, then verify it during deployment and roll back if it fails" arguably does, and it is smoke-testable — but it may be better as an extension to an existing API example than a new stack.

## Related

- #280 — `@composurecdk/custom-resources`, the adjacent escape hatch. The README distinction between the two mechanisms should be written as part of this work.

laazyj:
## What & why

Closes #365.

Some work has to happen *as part of shipping a stack*, and its outcome has to gate the deployment: registering a release with an external service, calling the deployed system to prove it answers, seeding reference data through a service API. Nothing in the library covered that, and **the nearest existing tool gets it silently wrong** — an `AwsCustomResource` running `Lambda.invoke` only reports `FAILED` when the *SDK call* throws, and a handler that throws still returns HTTP 200 with a `FunctionError` field, so the deploy goes green.

`.invokeOnDeploy()` invokes the function once during deployment and fails the deployment if the handler fails:

```ts
compose(
  {
    register: createFunctionBuilder()
      .runtime(Runtime.NODEJS_22_X)
      .handler("index.handler")
      .code(Code.fromAsset("register"))
      .timeout(Duration.seconds(30))
      .environment({ API_URL: "https://releases.example.com/v1/register" })
      .invokeOnDeploy({ after: [ref("api", (r: RestApiBuilderResult) => r.api)] }),

    api: createRestApiBuilder().restApiName("Orders"),
  },
  { api: [], register: ["api"] },
).build(stack, "Release");
```

### Shape: a domain action, not a standalone component

The issue left the shape open. This is a `<verb>()` on the function builder per [ADR-0016](https://github.com/laazyj/composureCDK/blob/main/docs/adr/0016-domain-action-custom-resource.md): "invoke this function during deployment" has no CloudFormation resource, and the builder that owns the Lambda domain owns the operation. The consumer declares intent; the builder owns the invocation semantics, the wait derivation, the execution-role ordering, and the IAM. The custom resource lands on the build result as `deploymentTrigger`, so a sibling can be ordered against the call itself (`deploymentTrigger.executeBefore(...)`).

It stays in `@composurecdk/lambda` rather than becoming a package: it uses only `aws-cdk-lib` types (`triggers` ships inside `aws-cdk-lib`), so it adds no dependency and no new-package machinery. `cdk-floors.json` is unchanged — `TriggerProps.invocationType`, `.timeout` and `.executeOnHandlerChange` all exist at lambda's declared floor of 2.168.0, verified by unpacking that version rather than assuming.

### Defaults and guards

| Behaviour | Default | Rationale |
| --- | --- | --- |
| Invocation type | `RequestResponse`, **fixed** | The deployment waits for the result. The asynchronous alternative returns before the work happens and would report success whatever the handler did — the opposite of the point. Not exposed as an option; raw `Trigger` remains the escape hatch. |
| Wait | function `timeout` + 30s, capped at 15 min | The deployment outlives the handler, so a handler that runs to its limit reports *its* error ("Task timed out after …") instead of the deployment abandoning the call and surfacing an unactionable provider timeout. The cap is the trigger provider's own Lambda timeout. Falls back to 2 min (CDK's default) when the function timeout is unset or a token. |
| `executeOnHandlerChange` | `true` | Re-invokes on a handler change, no-op on unrelated updates. |

A suppressible warning (`DEPLOY_INVOKE_TIMEOUT_WARNING_ID`) fires whenever the final wait is at or below the function's own timeout. It is checked against the *derived* wait too, not just an explicit one — at Lambda's 15-minute maximum the cap leaves no margin, which is exactly the state the guard exists to flag. Both operands belong to this builder and are final at `build()`, so it is a build-time warn like `warnIfLowMaxReceiveCount` in `@composurecdk/sqs`, not an ADR-0011 synth-time Aspect (which is for values owned by *another* component) and not an ADR-0010 throw (the configuration is deployable).

Ordering is data, not glue: the function's execution role is always waited for — CloudFormation does not otherwise sequence `.grant()`/`.configureRole()` policies ahead of a custom resource, and the handler would race them — and everything else goes in `after` as concrete constructs or `ref(...)` to siblings.

### Documentation

Both READMEs now document the trap that sinks this: **a handler that returns an error instead of throwing is a successful invocation and deploys green.** The `@composurecdk/custom-resources` README says why `Lambda.invoke` there is not the tool, since that is where a consumer looks first.

### ADR-0016 → Accepted

ADR-0016 was `Proposed` "until a second SDK-only operation is encapsulated this way. The second instance prices whether to extract shared scaffolding." This is that second instance, so the ADR is updated as it asked: status `Accepted`, no scaffolding extracted (both instances need different backings and share no plumbing), plus a third backing category — **an upstream `aws-cdk-lib` construct that already implements the operation** — which is what `triggers.Trigger` is here. Happy to drop that commit range from the PR if you would rather record it separately.

### Deliberately not included

- **No example stack.** The example bar wants a deployable, smoke-testable system; a deploy-time call needs a reachable endpoint at deploy time, and pointing CI at an external one is not something to bake into `deploy-test`. The natural in-repo version — call the API this stack just deployed — needs the API's URL in the handler's environment, and `FunctionBuilderProps.environment` is not `Resolvable`, so it cannot be wired under `compose` today. Worth a follow-up (widen `environment` to `Resolvable`), and then this earns an example that proves failure propagation end to end against live AWS.
- **`after` takes `Resolvable<IConstruct>[]`, not `Ref<object>[]`** like `.dependsOn()` in `@composurecdk/custom-resources`. Aligning them means promoting that package's `collectConstructs`/`addDependenciesFromRefs` into `@composurecdk/core` and widening its parameter — a change to two packages' published surface that does not belong in this PR. Follow-up if you want the two seams spelled the same way.

## Checklist

- [x] Linked to an issue (#365)
- [x] `npm run verify` passes locally — **except `actionlint`**, whose `shellcheck` binary download is blocked by this environment's network policy (403 from the proxy). No workflow files are touched by this PR. Every other gate in the chain was run individually and passes: `format:check`, `build`, `catalogue:check`, `check:exports`, `lint`, `cdk-floors:check`, `validate`, `test` (162 tests in `@composurecdk/lambda`, coverage thresholds met). The pre-push hook was bypassed for the same reason.
- [x] Tests added/updated for the change — 24 new tests covering failure propagation, wait derivation and the cap, the timeout guard (including suppression and token cases), execution-role and `after` ordering, and `[COPY_STATE]`
- [x] N/A — adds no example stack (see above)

https://claude.ai/code/session_013eSCRnprW4KKddjs7NQs5J

---
_Generated by [Claude Code](https://claude.ai/code/session_013eSCRnprW4KKddjs7NQs5J)_

### DIFF
--- a/docs/adr/0016-domain-action-custom-resource.md
+++ b/docs/adr/0016-domain-action-custom-resource.md
@@ -1,6 +1,6 @@
 # ADR 0016: Encapsulate SDK-only operations as domain actions on the owning builder
 
-- **Status:** Proposed
+- **Status:** Accepted
 - **Date:** 2026-07-12
 
 ## Context
@@ -26,5 +26,6 @@ The first application is `@composurecdk/ses` `.activate()`: a receipt rule set i
 
 - SDK-only operations a domain builder owns are expressed as domain actions, not consumer-assembled plumbing — encapsulated, in-domain, IAM-scoped, and surfaced on the build result.
 - The stateless-vs-stateful backing is implementation guidance, not part of the core decision: the same `.activate()` encapsulation would hold even if activation used an unconditional stateless `AwsCustomResource`.
-- **Proposed** until a second SDK-only operation is encapsulated this way. The second instance prices whether to extract shared scaffolding — a `domainActionProvider(scope, id, { handler, actions })` for the stateful case, and/or handler-serialisation helpers — rather than each builder rolling its own.
+- **Accepted.** The second instance is `@composurecdk/lambda` `.invokeOnDeploy()` (invoke this function during deployment, and fail the deployment if it fails — an operation with no CloudFormation resource, owned by the builder that owns the Lambda domain). It settles the pricing question the **Proposed** status was held open for: **no shared scaffolding is extracted yet.** Both instances so far need a different backing, and neither reuses the other's plumbing — so `domainActionProvider(...)` and handler-serialisation helpers stay unbuilt until two actions actually want the same one.
+- The backing categories are **three**, not two: alongside the stateless single call (case 1) and the bespoke `Provider` (case 2) sits **case 3 — an upstream `aws-cdk-lib` construct that already implements the operation** (here `triggers.Trigger`, which invokes synchronously and reports `FAILED` on a handler error). Prefer it when one exists: it is the least code and the semantics are maintained upstream. The domain action's job is then to express the operation in domain language, apply the defaults, and wire the ordering the raw construct leaves to the consumer.
 - Consumers running multiple instances of an account-global operation (e.g. several SES rule sets across stacks) must understand the shared-state arbitration; document it on the builder, and keep `createAwsCustomResourceBuilder` as the escape hatch for bespoke behaviour.
--- a/packages/custom-resources/README.md
+++ b/packages/custom-resources/README.md
@@ -8,6 +8,8 @@ This package wraps `AwsCustomResource` as a builder so those calls become first-
 
 > **When a domain builder already covers the call you need, prefer it** — it scopes IAM automatically and reads as intent rather than plumbing (e.g. a future SES `.activate()`). This builder is for the long tail of one-off SDK calls that don't justify a domain builder.
 
+> **Invoking your own Lambda during deployment is a different job** — use [`.invokeOnDeploy()`](../lambda/README.md#deploy-time-invocation-invokeondeploy) on the function builder. Reaching for `onCreate: { service: "Lambda", action: "invoke" }` here **will not fail the deployment when your handler fails**: the provider only reports `FAILED` when the _SDK call_ throws, and a handler that throws still returns HTTP 200 with a `FunctionError` field, so the deploy goes green.
+
 ## Usage
 
 ```ts
--- a/packages/lambda/README.md
+++ b/packages/lambda/README.md
@@ -379,6 +379,101 @@ today (the queue often arrives as an unresolved `ref()`); they are tracked in
 `kinesisEventSource` is still deferred — see
 [#120](https://github.com/laazyj/composureCDK/issues/120).
 
+## Deploy-time invocation: `.invokeOnDeploy()`
+
+Some work has to happen _as part of shipping the stack_, and its outcome has to
+gate the deployment: registering the release with an external service, calling
+the system you just deployed to prove it answers, seeding reference data through
+a service API. `.invokeOnDeploy()` invokes the function once during deployment
+and **fails the deployment if the function fails** — the stack rolls back and
+the handler's error message lands in the CloudFormation events.
+
+```ts
+import { compose, ref } from "@composurecdk/core";
+import { createFunctionBuilder } from "@composurecdk/lambda";
+
+compose(
+  {
+    register: createFunctionBuilder()
+      .runtime(Runtime.NODEJS_22_X)
+      .handler("index.handler")
+      .code(Code.fromAsset("register"))
+      .timeout(Duration.seconds(30))
+      .environment({ API_URL: "https://releases.example.com/v1/register" })
+      .invokeOnDeploy({ after: [ref("api", (r: RestApiBuilderResult) => r.api)] }),
+
+    api: createRestApiBuilder().restApiName("Orders"),
+  },
+  { api: [], register: ["api"] },
+).build(stack, "Release");
+```
+
+It is a domain action on the function builder ([ADR-0016](../../docs/adr/0016-domain-action-custom-resource.md)),
+backed by the CDK [`Trigger`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.triggers-readme.html)
+construct. The custom resource is exposed on the build result as
+`deploymentTrigger`, so a sibling can be ordered against the call itself
+(`deploymentTrigger.executeBefore(...)`).
+
+### The handler must throw
+
+**A handler that returns an error rather than throwing does not fail the
+deployment.** Lambda considers an invocation that returns successful no matter
+what the payload says, so an HTTP-shaped `{ statusCode: 500 }` — or a caught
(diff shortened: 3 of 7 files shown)


laazyj:
Currently refactoring and updating...

PR #375, opened 2026-08-13
feat(kms): add @composurecdk/kms and widen key-consuming props to Resolvable

### LINKED ISSUE #373: feat(kms): add @composurecdk/kms KeyBuilder and resolvable encryption-key props so CMKs can be compose components
**Problem / use case**

A system whose resources are encrypted with customer-managed KMS keys can't express the keys as components, because there is no KMS builder and no key-consuming prop accepts a `Ref`. The keys have to be constructed imperatively, outside `compose`, and handed to the builders as concrete constructs.

Real example — a service with three DynamoDB tables and an S3 dead-letter bucket, all CMK-encrypted (`@composurecdk/dynamodb@0.9.2`, `@composurecdk/s3@0.9.2`, `@composurecdk/core@0.9.2`):

```ts
// Four constructs that have to be created by hand, before compose, with their
// stack chosen manually — the only part of this system that isn't a component.
const taskStateTableKey = new Key(persistenceStack, "TaskStateTableKey", { enableKeyRotation: true });
const taskConfigTableKey = new Key(persistenceStack, "TaskConfigTableKey", { enableKeyRotation: true });
const insightMappingTableKey = new Key(persistenceStack, "InsightMappingTableKey", { enableKeyRotation: true });
const taskPublisherDlqBucketKey = new Key(persistenceStack, "TaskPublisherDlqBucketKey", { enableKeyRotation: true });

compose(
  {
    taskStateTable: createTableV2Builder()
      .partitionKey({ name: "taskId", type: AttributeType.STRING })
      .dynamoStream(StreamViewType.NEW_AND_OLD_IMAGES)
      .encryption(TableEncryptionV2.customerManagedKey(taskStateTableKey)),   // closed-over construct
    // …taskConfigTable, insightMappingTable, same shape…

    taskPublisherDlqBucket: createBucketBuilder()
      .encryption(BucketEncryption.KMS)
      .encryptionKey(taskPublisherDlqBucketKey),                              // closed-over construct
  },
  {
    taskStateTable: [],            // <- the key edges are invisible here
    taskPublisherDlqBucket: [],
    // …
  },
).withStacks({
  taskStateTable: persistenceStack,
  taskPublisherDlqBucket: persistenceStack,
  // …the keys aren't listed; their scope was hardcoded above
});
```

The friction, in order of how much it costs:

1. **The dependency map stops being the whole topology.** `taskStateTable` reads as having no dependencies when it has one — its key. The component/dependency maps are the thing you read to understand the system, and four real edges are missing from them.
2. **Stack placement is manual and can silently diverge.** Every component's scope comes from `.withStacks()`; the keys' scope comes from the `new Key(persistenceStack, …)` argument. Move the tables to another stack and the keys stay behind, with nothing to catch it — no `withStacks` entry to update, and none of the cross-stack relationship guards (ADR-0011) apply.
3. **No graph services for keys.** They're outside the build order, cycle detection, construct-id policy (ADR-0012), and tagging that every other resource gets.
4. **The imperative prologue undoes the point of the declarative form.** The system is otherwise fully declarative; secure-by-default encryption is exactly the case you'd want kept in the graph, and it's the one that has to escape it.

**Proposed solution**

Two parts. Together they let a CMK be an ordinary component that its consumers depend on by name.

**1. New `@composurecdk/kms` package with a `KeyBuilder`**

Standard builder shape and secure defaults (key rotation on, `RemovalPolicy.RETAIN` — losing a key means losing the data it protects):

```ts
import { createKeyBuilder, KeyBuilderResult, keyGrants } from "@composurecdk/kms";

taskStateTableKey: createKeyBuilder()
  .description("Encrypts the Task State Table at rest.")
  .alias("promote-tasks/task-state"),   // .enableKeyRotation(true) by default
```

`KeyBuilderResult` would carry `{ key: Key }`. Consumer-side grants per ADR-0013 for the cases a table/bucket grant doesn't already cover — e.g. a Lambda that talks to a CMK-encrypted resource it doesn't hold a resource grant on:

```ts
export const keyGrants = {
  encrypt: (key: Resolvable<IKey>) => Grant<IGrantable>,
  decrypt: (key: Resolvable<IKey>) => Grant<IGrantable>,
  encryptDecrypt: (key: Resolvable<IKey>) => Grant<IGrantable>,
  admin: (key: Resolvable<IKey>) => Grant<IGrantable>,
};
```

**2. Widen the key-consuming props to `Resolvable`**

Same three-step widening `DistributionBuilder.certificate`, `RoleBuilder.assumedBy`, `FunctionBuilder.role` and (as of #356) `SpecRestApiBuilder.apiDefinition` already use — `Omit<CdkProps, "x">` + a `Resolvable` re-declaration, resolved in `build()`:

- `@composurecdk/dynamodb` — `TableV2BuilderProps.encryption?: Resolvable<TableEncryptionV2>` (and the classic `TableBuilder`'s `encryption`/`encryptionKey`)
- `@composurecdk/s3` — `BucketBuilderProps.encryptionKey?: Resolvable<IKey>`

The same shape exists on every other builder that takes a key (SQS `encryptionMasterKey`, SNS `masterKey`, Logs `encryptionKey`), so it's worth deciding once whether to widen those in the same pass or as follow-ups.

The example above then becomes fully declarative — no prologue, and the key edges are in the map:

```ts
compose(
  {
    taskStateTableKey: createKeyBuilder().description("Encrypts the Task State Table at rest."),
    taskPublisherDlqBucketKey: createKeyBuilder().description("Encrypts the task publisher DLQ bucket."),

    taskStateTable: createTableV2Builder()
      .partitionKey({ name: "taskId", type: AttributeType.STRING })
      .dynamoStream(StreamViewType.NEW_AND_OLD_IMAGES)
      .encryption(ref("taskStateTableKey", (r: KeyBuilderResult) => TableEncryptionV2.customerManagedKey(r.key))),

    taskPublisherDlqBucket: createBucketBuilder()
      .encryption(BucketEncryption.KMS)
      .encryptionKey(ref<KeyBuilderResult>("taskPublisherDlqBucketKey").get("key")),
  },
  {
    taskStateTableKey: [],
    taskPublisherDlqBucketKey: [],
    taskStateTable: ["taskStateTableKey"],
    taskPublisherDlqBucket: ["taskPublisherDlqBucketKey"],
  },
).withStacks({
  taskStateTableKey: persistenceStack,
  taskPublisherDlqBucketKey: persistenceStack,
  taskStateTable: persistenceStack,
  taskPublisherDlqBucket: persistenceStack,
});
```

One ergonomic question for the DynamoDB side: `TableEncryptionV2.customerManagedKey(key)` is a static factory over the key, so the `ref` mapping is a little indirect. A `Resolvable<TableEncryptionV2>` keeps the CDK prop surface unchanged and handles every encryption mode uniformly (`awsManagedKey`, `dynamoOwnedKey`, replica key maps); a narrower `.encryptionKey(Resolvable<IKey>)` convenience that implies `customerManagedKey` would read better at the call site but adds a second, mutually-exclusive way to say the same thing (ADR-0009 territory). Happy either way — flagging it as the part worth agreeing on before implementation.

**Alternatives considered**

- **Wrap `Key` in a hand-written `Lifecycle` adapter locally.** This works today and is what a stop-gap for #356 looked like for the spec REST API, but a `Lifecycle` per key is boilerplate in every consuming repo for something as common as a CMK, and it still can't be referenced by the widened props — a table would need the adapter's result threaded in via context by hand.
- **Keep the keys imperative and just add them to `.withStacks()`.** Not possible: `withStacks` keys off component names, and a raw construct isn't a component. There's no way to get a hand-built `Key` into the graph at all.
- **Use AWS-managed keys and avoid the problem.** Not an option when you need control over the key policy, CloudTrail visibility of decrypt calls, and independent rotation — which is the reason for CMKs in the first place, and the current builder default (`TableEncryptionV2.awsManagedKey()`) is what's being opted out of.
- **Do the widening only, no `kms` package.** Halfway: `Resolvable` props would have nothing to reference, since a component has to be a `Lifecycle`. The builder is the part that makes keys first-class; the widening is what makes them reachable.

Prior art: CDK's own `Key` construct is the thing being wrapped, and `@composurecdk/acm`'s `CertificateBuilder` + `DistributionBuilder.certificate` is the closest existing pair in this repo — a small "supporting resource" builder plus a widened prop on its consumer. This is the same pattern for encryption keys.

Happy to take this on as two PRs (package, then widening) if the approach looks right.

laazyj:
## What & why

Closes #373.

A system encrypted with customer-managed KMS keys couldn't express those keys as components. The friction the issue describes is real and it isn't the boilerplate — it's that the dependency map stops being the whole topology, and that stack placement for the keys comes from a `new Key(stack, …)` argument rather than `.withStacks()`, free to diverge from the resources they protect with nothing to catch it.

Two parts.

### 1. `@composurecdk/kms`

`createKeyBuilder()` in the standard builder shape — `taggedBuilder`, `addAlarm`/`[COPY_STATE]`, `KEY_DEFAULTS` exported.

| Default | Value | Rationale |
| --- | --- | --- |
| `enableKeyRotation` | `true` | Transparent to existing ciphertext ([Security Pillar](https://docs.aws.amazon.com/wellarchitected/latest/security-pillar/sec_protect_data_rest_key_mgmt.html)) |
| `removalPolicy` | `RETAIN` | Deleting a key destroys every ciphertext it protects |
| `pendingWindow` | `Duration.days(30)` | The maximum window in which a scheduled deletion can still be cancelled |

AWS only rotates symmetric encryption keys, so the rotation default yields to an asymmetric or HMAC `keySpec` rather than letting CDK reject the pair (ADR-0009); asking for both explicitly still throws. `KeyBuilderResult` carries `{ key, alias?, alarms }` — the alias is created via `addAlias` rather than passed as `KeyProps.alias`, so the construct is visible in the result under the same logical id.

`keyGrants` provides consumer-side grants (ADR-0013) for what a resource grant can't infer: a principal decrypting ciphertext fetched elsewhere, an envelope-encryption client calling `GenerateDataKey`, or a key shared with a resource the grantee holds no grant on. `admin` delegates to `Key.grantAdmin` and so takes a concrete `Key` — the alternative, the library carrying its own copy of AWS's admin action set, is what ADR-0013's addendum rules out.

**Recommended alarms.** The one AWS-recommended KMS alarm covers `SecondsUntilKeyMaterialExpiration`, and it is opt-in rather than on by default — the only package where that's so. KMS publishes the metric solely for keys with imported, expiring key material, so on a CloudFormation-created key the alarm could never leave `INSUFFICIENT_DATA`, and an alarm that can't fire is noise. `treatMissingData: ignore` latches it: KMS stops publishing once the material has actually expired, and `notBreaching` would clear the alarm at exactly the moment the key stopped working.

### 2. Widened key-consuming props

The `Omit` + `Resolvable` re-declaration + resolve-in-build pattern already used by `DistributionBuilder.certificate` and `VolumeBuilder.encryptionKey`:

| Package | Prop | Accepts |
| --- | --- | --- |
| `s3` | `encryptionKey` | `Resolvable&lt;IKey&gt;` |
| `dynamodb` | `encryption` (TableV2) | `Resolvable&lt;TableEncryptionV2&gt;` |
| `dynamodb` | `encryptionKey` (classic `Table`) | `Resolvable&lt;IKey&gt;` |
| `sqs` | `encryptionMasterKey` | `Resolvable&lt;IKey&gt;` |
| `sns` | `masterKey` | `Resolvable&lt;IKey&gt;` |
| `logs` | `encryptionKey` | `Resolvable&lt;IKey&gt;` |

Where the library's own encryption default is mutually exclusive with a customer key, supplying one infers the customer-managed mode (ADR-0009) — S3 `S3_MANAGED` → `KMS`, SQS `SQS_MANAGED` → `KMS` — matching what the classic table builder already did for `CUSTOMER_MANAGED`.

On the DynamoDB ergonomic question the issue raised: the `Resolvable` sits on the whole `TableEncryptionV2`. It keeps the CDK prop surface unchanged, handles every mode uniformly including per-replica key ARNs, and adds no second mutually-exclusive spelling of the same thing.

## CDK floor

Established at **2.93.0** and proven with `cdk-floors enforce` — the lowest reachable, since the package peers `@composurecdk/cloudwatch` for its alarms. Every aws-kms API used sits far below it, with one exception recorded in the manifest: `IKey.grantSign` / `grantVerify` / `grantSignVerify` do **not** exist at 2.93.0, so those two capabilities name `kms:Sign` / `kms:Verify` through `IKey.grant` instead of delegating. The `gatedBy` note says so explicitly, since the obvious "simplification" there doesn't compile at the floor. The five widened packages' floors were re-verified unchanged (2.1.0, 2.93.0, 2.123.0, 2.178.0).

## Deliberately not included

- **An example stack.** A bare CMK doesn't clear the bar in AGENTS.md (a system, not a resource), and CI deploys every example — a RETAIN-by-default key would outlive the sandbox teardown.
- **`lambda`'s `environmentEncryption` and `neptune`'s `kmsKey`.** Both are typed `IKeyRef` on current CDK but `IKey` at their floors, so widening them narrows the accepted type for consumers on newer CDK. Left concrete and called out in the KMS README; worth deciding separately.
- **An ADR.** This applies existing patterns to a new service. But flagging one thing for a follow-up: **ADR-0009's own escalation clause has fired.** This adds three more default-vs-sibling yields, in three different shapes (an extracted helper in s3/sqs, an inline assignment in dynamodb, an inline delete in kms) — the recurrence the ADR named as the signal to weigh a typed primitive in `core`. That binds work across packages, so it wants its own ADR rather than being smuggled in here.

## Checklist

- [x] Linked to an issue (#373)
- [x] `npm run verify` passes locally — with one exception: `actionlint` cannot run in the authoring environment, as its egress proxy blocks the shellcheck binary download (403). No workflow files are touched by this change; every other gate in the chain (`format:check`, `build`, `catalogue:check`, `check:exports`, `lint`, `cdk-floors:check`, `validate`, `test`) was run and passes, and CI has since run the full chain green on Node 20/22/24/26.
- [x] Tests added/updated — new suites for the KMS builder, alarms and grants at 100% coverage, plus resolve-from-context and mode-inference tests in each widened package
- [x] No example stack added (see above)


packages/kms/src/alarm-config.ts
```diff
@@ -0,0 +1,41 @@
+import type { AlarmConfig } from "@composurecdk/cloudwatch";
+
+/**
+ * Controls which recommended alarms are created for a KMS key.
+ *
+ * Unlike the other builder packages, the single AWS-recommended KMS alarm is
```
laazyj: Remove the "Unlike the other builder packages" qualification. This may not hold true indefinitely and could be confusing for consumers. Just explain that this alarm is Opt-in, why, and the circumstances when you should opt-in.

packages/kms/src/grants.ts
```diff
@@ -0,0 +1,87 @@
...
+  verifyMac: capability((key, grantee) => {
+    key.grantVerifyMac(grantee);
+  }),
+  /**
+   * Administer the key — describe, tag, enable/disable, schedule deletion —
+   * without any permission to use it cryptographically. Reserve it for an
+   * operator or automation role; workloads want {@link keyGrants.encryptDecrypt}.
+   *
+   * Takes a concrete {@link Key} rather than an {@link IKey}: `grantAdmin` is
+   * declared on the class, and the alternative — the library carrying its own
```
laazyj: drop the "and the alternative" clause. We don't need to re-litigate our design decisions in the API documentation.

laazyj (commented):
Some documentation simplification comments.

I would also like to see a follow-up PR that migrates an existing example app (`crud-api-app.ts`) to use a customer-managed KMS key. This will prove its capabilities. To minimise costs the `clean-desk-policy` should be updated to delete keys and immediately expire deleted keys.

packages/kms/src/alarm-config.ts
laazyj: Done in cca544d — the doc now states the fact, the reason, and when to opt in, with no cross-package comparison.

I applied the same reasoning to two places you didn't mark, since they carried the identical framing: `KEY_ALARM_DEFAULTS`'s doc ("defaults on as it does in every other package") and the README's Recommended Alarms section ("Unlike every other builder in the library"). Say the word if you'd rather I left those.

---

packages/kms/src/grants.ts
laazyj: Done in cca544d. Dropped the clause; kept the typing constraint, since "takes a `Key`, not an `IKey`" is something a caller hits at the call site and needs the workaround for.

---

packages/kms/README.md
```diff
@@ -0,0 +1,205 @@
...
+| Package                  | Prop                  | Accepts                              |
+| ------------------------ | --------------------- | ------------------------------------ |
+| `@composurecdk/s3`       | `encryptionKey`       | `Resolvable<IKey>`                   |
+| `@composurecdk/dynamodb` | `encryption` (V2)     | `Resolvable<TableEncryptionV2>`      |
+| `@composurecdk/dynamodb` | `encryptionKey`       | `Resolvable<IKey>` (classic `Table`) |
+| `@composurecdk/sqs`      | `encryptionMasterKey` | `Resolvable<IKey>`                   |
+| `@composurecdk/sns`      | `masterKey`           | `Resolvable<IKey>`                   |
+| `@composurecdk/logs`     | `encryptionKey`       | `Resolvable<IKey>`                   |
+
+Two key-consuming props elsewhere still take a concrete key — `@composurecdk/lambda`'s `environmentEncryption` and `@composurecdk/neptune`'s `kmsKey`. Build the key as a component and pass `result.key` to those until they are widened.
```
laazyj: Raise GitHub issues to track these widening tasks for KeyBuilder support (`lambda`, `neptune`)

laazyj (commented):
2 GitHub issues required to track follow-ups

packages/kms/README.md
laazyj: Raised #379 (lambda `environmentEncryption`) and #380 (neptune `kmsKey`), and linked both from this paragraph in 9583039.

Each issue carries the `IKeyRef` decision rather than assuming it, since that's the part that isn't mechanical: both props are typed `kms.IKeyRef` on current CDK but `IKey` at their floors, so `Resolvable<IKey>` compiles at the floor while narrowing what a newer-CDK consumer can pass, and `Resolvable<IKeyRef>` does the reverse. Neptune has the extra wrinkle that its alpha peer is version-locked, so raising that floor moves both.

---

PR #381, opened 2026-08-15
feat(route53): add a CrossAccountZoneDelegationRecord builder

### LINKED ISSUE #378: feat(route53): add a CrossAccountZoneDelegationRecord builder
## Problem / use case

#376 added `hostedZoneGrants.delegation(...)`, which covers the **parent** account's half of a cross-account subdomain delegation: a role the child account may assume to write NS records into the parent zone, optionally confined to named subdomains.

The **child** account's half has no builder. Once the child owns `beta.example.com` and holds the delegation role's ARN, it still has to publish its own zone's name servers into the parent zone, and the only construct for that — `CrossAccountZoneDelegationRecord` — has no fluent equivalent. So a system that gets the grant right declaratively still drops into a hand-written `Lifecycle` for the last step, exactly the wrapper the grant work removed:

```ts
class DelegationRecordLifecycle implements Lifecycle {
  build(scope: IConstruct, id: string, context: Record<string, object> = {}) {
    const { hostedZone } = resolve(this.childZoneRef, context);
    return new CrossAccountZoneDelegationRecord(scope, id, {
      delegatedZone: hostedZone,
      parentHostedZoneName: "example.com",
      delegationRole: Role.fromRoleArn(scope, "ParentRole", this.roleArn),
    });
  }
}
```

The two halves are one workflow, and the package now ships one of them. `packages/route53/README.md` says so explicitly at the end of the "Delegation grants" section:

> Publishing the delegated zone's own `NS` records into a parent zone owned by **another** account is the other half of this story, and has no builder yet — use CDK's `CrossAccountZoneDelegationRecord` with the role granted above.

**That paragraph must be replaced (not merely appended to) when this lands**, and the delegation-grants section should then read as one end-to-end story: grant on the parent, record on the child.

## Proposed solution

`createCrossAccountZoneDelegationBuilder()`, following the existing record builders' shape (fluent setter per CDK prop, `Resolvable<T>` on anything that can come from a sibling, `ITaggedBuilder`, a result object keyed by role).

```ts
compose(
  {
    childZone: createHostedZoneBuilder().zoneName("beta.example.com"),

    parentDelegation: createCrossAccountZoneDelegationBuilder()
      .delegatedZone(ref("childZone", (r: HostedZoneBuilderResult) => r.hostedZone))
      .parentHostedZoneName("example.com")
      .delegationRole(parentDelegationRoleArn) // the role #376 granted, in the parent account
      .ttl(Duration.minutes(30)),
  },
  { childZone: [], parentDelegation: ["childZone"] },
);
```

`CrossAccountZoneDelegationRecordProps` in full, for reference — `delegatedZone`, `parentHostedZoneName`, `parentHostedZoneId`, `delegationRole`, `ttl` (default `Duration.days(2)`), `removalPolicy` (default `DESTROY`), `assumeRoleRegion`.

Notes on the surface:

- **`delegationRole` should accept a plain role ARN string as well as `Resolvable<IRole>`.** In the real use case the role lives in another account, so the child stack has nothing to `ref` — it has an ARN, and today every call site writes the same `Role.fromRoleArn(...)` boilerplate. Accepting the ARN and importing internally removes it; a `Resolvable<IRole>` still covers the same-account and same-app cases.
- **`parentHostedZoneName` / `parentHostedZoneId` are mutually exclusive and one is required.** CDK throws at construction; the builder should surface it as a build-time error naming both setters, per [ADR-0010](https://github.com/laazyj/composureCDK/blob/main/docs/adr/0010-aws-property-constraints.md), with [ADR-0009](https://github.com/laazyj/composureCDK/blob/main/docs/adr/0009-defaults-yield-to-mutually-exclusive-siblings.md) governing any default either one might otherwise get.
- **Guard the zone-flavour constraint.** The construct passes `delegatedZone.hostedZoneNameServers` to the custom resource, and that attribute is `undefined` for private hosted zones and for zones imported via `fromHostedZoneAttributes` / across stacks. Passing one of those produces a delegation record with no name servers rather than an error. This is the [ADR-0011](https://github.com/laazyj/composureCDK/blob/main/docs/adr/0011-cross-component-relationship-guards.md) case: fail at build with a message that names the zone and the remediation.

### Well-Architected considerations

**Security.** The construct creates a Lambda-backed custom resource (`Custom::CrossAccountZoneDelegation`, one provider singleton per stack) and grants that provider's role `sts:AssumeRole` on precisely the one delegation-role ARN passed in — already least privilege, so the builder should not widen it. The remaining lever is on the parent side: pair this with `hostedZoneGrants.delegation(zone, { delegatedZoneNames: [...] })` so the assumed role can only touch this subdomain. The README example should show both halves together with the scoping in place, since the unscoped grant is the tempting shortcut.

**Operational excellence.** The provider Lambda's log group is created by aws-cdk-lib with no retention — it never expires and is invisible to `@composurecdk/logs` defaults, unlike every other log group the library provisions (the hosted zone's query-log group already applies `RetentionDays.TWO_YEARS` / `RemovalPolicy.RETAIN`). The builder should bring the provider's log group under the same defaults, with the same `configure` escape hatch the query-logging config uses. Worth confirming during design how much of the singleton provider is reachable, since it is created via `getOrCreateProvider`.

**Reliability.** Two defaults carry real consequences and should be documented rather than inherited silently:

- `ttl` defaults to 2 days. That is fine steady-state and painful during a migration — an NS change takes up to the old TTL to propagate. Recommend documenting a lower TTL (minutes) ahead of a planned delegation change, then raising it again.
- `removalPolicy` defaults to `DESTROY`, which removes the NS records from the parent zone when the child stack is deleted. That is the correct hygiene default (`RETAIN` leaves a lame delegation pointing at a zone that no longer exists) and should stay, with the trade-off spelled out — the records are in *someone else's* account, so a stack deletion reaches across an account boundary.

Also worth documenting: the record is written at deploy time only. If the child zone is ever replaced, its name servers change and the parent's NS records go stale until the next deploy; nothing detects that drift.

**Cost.** Negligible — one custom-resource invocation per deploy, plus the log group above.

### Recommended alarms

Being straight about this: **there is nothing to alarm on for the record itself.** Route 53 publishes no CloudWatch metrics for record sets, and the custom resource runs only at deploy time, so alarms on the provider Lambda would be noise rather than signal — a deploy-time failure already fails the deployment. The AWS-recommended alarms that *are* relevant to a delegation are one layer out:

- **`HealthCheckStatus`** — AWS's recommended Route 53 alarm, and the practical liveness signal for a delegation, because a health check against an endpoint in the delegated zone only stays healthy while resolution works end to end through the parent's NS records. The package already has this: `createHealthCheckBuilder` + `createHealthCheckAlarmBuilder` (`packages/route53/src/health-check-alarms.ts`, fixed at the AWS-recommended `Minimum` / 1-minute basis). The README section should point at it rather than the builder auto-creating one — a health check needs an endpoint the builder has no way to know.
- **DNSSEC** — if the delegated zone is signed, the delegation's DS half is what breaks, and AWS recommends alarming on `DNSSECInternalFailure` and `DNSSECKeySigningKeysNeedingAction`. The package has no zone-level alarms today (only health-check alarms), so this is a separate gap; I'd rather file it than smuggle it into this builder. Happy to raise it if useful.

So the recommendation for this issue is: **no auto-provisioned alarms**, and a README subsection explaining what to alarm on instead and why the record itself is not it.

### Delivery checklist

- `packages/route53/src/cross-account-zone-delegation-builder.ts` + export from `src/index.ts`
- Defaults recorded in `src/defaults.ts` alongside the other record defaults, if any are applied
- `packages/route53/test/` synth test covering: the happy path, the name/id mutual exclusion, the missing-name-servers guard, ARN-vs-`IRole` delegation role, and TTL/removal-policy defaults (route53 is held to the 90% branch floor from #261)
- README: replace the "no builder yet" paragraph, add the end-to-end grant + record example, document the alarm guidance above
- No new example stack: exercising this needs a second AWS account, so it cannot carry a smoke test that proves anything beyond "the stack synthesised" — it fails the examples bar in AGENTS.md
- No ADR expected: this is a new resource builder following existing patterns, not an architectural decision

## Alternatives considered

- **Leave it to a hand-written `Lifecycle`.** Status quo. It works, but it puts `resolve(...)` in application code and reintroduces the wrapper #376 was written to remove — for the second half of the very same workflow.
- **Fold it into `hostedZoneGrants`.** Wrong direction and wrong account: the grant is declared by the parent on a grantee, the record is a resource the child creates. They are two constructs in two accounts that happen to meet at one role ARN.
- **Wait for an L2 with a nicer surface.** `CrossAccountZoneDelegationRecord` has been stable for years and is the only supported path; there is nothing pending to wait for.

laazyj:
## What & why

Closes #378.

`hostedZoneGrants.delegation(...)` (#376) covered the **parent** account's half of a cross-account subdomain delegation. This adds the **child**'s half — `createCrossAccountZoneDelegationBuilder()` — so the workflow no longer drops into a hand-written `Lifecycle` for its last step.

```ts
compose(
  {
    childZone: createHostedZoneBuilder().zoneName("beta.example.com"),

    parentDelegation: createCrossAccountZoneDelegationBuilder()
      .delegatedZone(ref("childZone", (r: HostedZoneBuilderResult) => r.hostedZone))
      .parentHostedZoneName("example.com")
      .delegationRole("arn:aws:iam::111122223333:role/delegation-beta")
      .ttl(Duration.minutes(30)),
  },
  { childZone: [], parentDelegation: ["childZone"] },
);
```

### Surface

- **`delegationRole` takes a plain ARN**, an `IRole`, or a ref to either (`Resolvable<string | IRole>`). The role lives in the parent account, so the child stack has nothing to `ref` — accepting the ARN removes the `Role.fromRoleArn(...)` line every call site would repeat. Imported with `mutable: false`: this stack assumes the role, it never attaches policies to it.
- **`parentHostedZoneName` / `parentHostedZoneId` are mutually exclusive, one required.** CDK throws at construction; the builder surfaces it at build naming both setters (ADR-0010). Neither is defaulted, so ADR-0009 does not come into play — noted in `defaults.ts` so the next reader doesn't have to re-derive it.
- **Zone-flavour guard.** The construct forwards `delegatedZone.hostedZoneNameServers` to the custom resource, and that attribute is `undefined` for private zones and for zones imported via `fromLookup`/`fromHostedZoneAttributes` or referenced across stacks — CDK would deploy a delegation with an empty `NS` set, resolving nowhere, with no error. The builder fails at build naming the zone and the remediation.

### Well-Architected

**Security.** CDK already grants the provider `sts:AssumeRole` on exactly the one ARN passed in, so there was nothing to tighten here and the builder does not widen it. The README makes the point that the lever that matters is `delegatedZoneNames` on the parent's grant, and links to it rather than restating it.

**Operational excellence.** aws-cdk-lib creates the provider as a raw `AWS::Lambda::Function` with no `LoggingConfig`, so the Lambda service implicitly creates a never-expiring log group that no template describes. The builder declares one under the `@composurecdk/logs` defaults and points `LoggingConfig` at it, with the same `configure` escape hatch `queryLogging` uses.

Two consequences worth calling out for review:

- **The name is pinned under `/aws/lambda/`.** CDK owns the provider's execution role and gives it only `AWSLambdaBasicExecutionRole`, which permits `logs:PutLogEvents` on `/aws/lambda/*` and nothing else — a log group named outside that prefix deploys fine and silently captures nothing. A `configure` that moves it warns.
- **The provider is a stack singleton, so its log group is too.** That makes `providerLogging` a per-record knob over a stack-wide resource. The dedup check therefore runs *before* the `false` opt-out, so the outcome doesn't depend on `compose`'s build order, and a later record whose setting couldn't be honoured is warned rather than silently ignored — including `providerLogging(false)`, which cannot remove a group a sibling already created.

**Reliability.** `ttl` (2 days) and `removalPolicy` (`DESTROY`) match CDK but are set explicitly so both land in the package's defaults table, per the `HEALTH_CHECK_DEFAULTS` precedent. The README documents lowering the TTL ahead of a planned delegation change, that `DESTROY` reaches across an account boundary to delete records in someone else's zone, and that the record is written at deploy time only — a replaced child zone leaves the parent's `NS` records stale until the next deploy, with nothing detecting the drift.

**Alarms: none provisioned**, matching the issue's analysis. Route 53 publishes no metrics for record sets, and the custom resource runs only at deploy time where a failure already fails the deployment. The README points at the health-check alarms this package already ships as the signal that actually catches a broken delegation, and notes zone-level DNSSEC alarms as a separate gap rather than smuggling them in here.

### Deviations from the issue, for your call

- **`IBuilder`, not `ITaggedBuilder`.** The record renders an `AWS::CloudFormation::CustomResource`, which has no `Tags` property — the case `builder-must-be-tagged` documents an escape for, with the same `eslint-disable` shape `NsRecordBuilder` uses. The only taggable resource in the result is the stack-shared provider log group, where per-record tags would be wrong.
- **`providerLogGroup` is exposed on the result**, unlike query logging's stack-shared `ResourcePolicy`. Architecture's "if a builder creates it, the result should expose it" won out for a log group, which is routinely wired to metric and subscription filters; the JSDoc and README say plainly that it is stack-shared.

### Files

- `src/cross-account-zone-delegation-builder.ts` — the builder, exported from `src/index.ts`
- `src/cross-account-delegation-provider-logging.ts` — provider log group + `DelegationProviderLoggingConfig`, split out the way `query-logging.ts` is
- `src/defaults.ts` — `CROSS_ACCOUNT_ZONE_DELEGATION_DEFAULTS` and the log-group constants
- `test/cross-account-zone-delegation-builder.test.ts` — 26 tests: happy path, name/id mutual exclusion, missing-name-servers guard, ARN vs `IRole` vs ref, TTL/removal-policy defaults and overrides, and the provider-logging singleton behaviour
- `README.md` — the "no builder yet" paragraph is replaced, the delegation story now reads end to end (grant on the parent, record on the child), and the alarm guidance is documented

No new example stack: exercising this needs a second AWS account, so it could not carry a smoke test proving anything beyond "the stack synthesised" — it fails the examples bar in AGENTS.md. No ADR: this applies existing patterns to a new resource.

## Checklist

- [x] Linked to an issue (Closes #378)
- [x] `npm run verify` passes locally — with one caveat: its `actionlint` step cannot run in my sandbox, because the `shellcheck` npm package downloads its binary from `github.com`, which this session's egress policy blocks (403). Every other step in the chain was run and passes: `format:check`, `build`, `catalogue:check`, `check:exports`, `lint`, `cdk-floors:check`, `validate`, `test`. No workflow files are touched by this PR, so `actionlint` has nothing to say about it — but it does mean the pre-push hook was skipped and CI is the first real run of that gate.
- [x] Tests added/updated for the change — route53 is at 169 tests, coverage above the package's 90% branch / 100% function floors
- [ ] If it adds an example stack — n/a, see above

---
_Generated by [Claude Code](https://claude.ai/code/session_01XdEfWKrsSjeqvMAugZTvAv)_


packages/route53/src/cross-account-delegation-provider-logging.ts
```diff
@@ -0,0 +1,179 @@
...
+  cfg: DelegationProviderLoggingConfig | undefined,
+): LogGroup | undefined {
+  const stack = Stack.of(scope);
+
+  // Ahead of the `false` check: an earlier record in this stack has already
+  // settled the singleton, and this record cannot unsettle it. Returning the
+  // group it will actually log to beats returning `undefined` and implying an
+  // isolation that does not exist.
+  const existing = stack.node.tryFindChild(DELEGATION_PROVIDER_LOG_GROUP_ID);
+  if (existing instanceof LogGroup) {
```
laazyj: Is using `instanceOf` going to be fragile to incidental changes in the definition of `LogGroup` across `aws-cdk-lib` versions? And this might then lead to unnecessary additional log groups?

Is there a more robust/defensive way to confirm the identified Node is a `LogGroup`?

packages/route53/src/defaults.ts
```diff
@@ -183,3 +184,62 @@ export const HEALTH_CHECK_DEFAULTS: Partial<HealthCheckBuilderProps> = {
...
+ */
+export const DELEGATION_PROVIDER_LOG_GROUP_NAME_PREFIX = "/aws/lambda";
+
+/**
+ * Construct id of the log group materialised once per stack for the
+ * cross-account delegation provider. The provider is a CDK stack singleton, so
+ * its log group is one too. Package-internal — used by the dedup helper, and
+ * intentionally not re-exported from the package barrel.
+ */
+export const DELEGATION_PROVIDER_LOG_GROUP_ID = "ComposureCDKRoute53DelegationProviderLogs";
```
laazyj: `DELEGATION_PROVIDER_LOG_GROUP_ID` and `DELEGATION_PROVIDER_LOG_GROUP_NAME_PREFIX` are both internal values that don't qualify as "default configuration" values - users cannot override them.

Tthey don't even need to know they exist since the values are used solely to derive configuration for other constructs, which are then exposed to the consumer via the "Result".

Recommend relocating these consts closer to their use sites and removing from public api.

laazyj (commented):
Question about type matching and a relocation of private constants.

packages/route53/src/cross-account-delegation-provider-logging.ts
laazyj: You're right, and the failure mode is worse than incidental version drift — it's the dual-package hazard we ship for. Fixed in 2c267ce.

The realm case is the sharp one. With the ESM and CommonJS copies of `aws-cdk-lib` both loaded in a process, the `LogGroup` class an earlier record constructed can be a different class object from the one this module imported, so `instanceof` returns `false` for a log group that is plainly sitting at that construct id.

On the "additional log groups" question — it doesn't get that far, but it doesn't degrade gracefully either. A missed dedup falls through to `subBuilder.build(stack, DELEGATION_PROVIDER_LOG_GROUP_ID)`, which collides with the existing child and throws a duplicate-construct-id error. So the symptom is a hard synth failure, not a silent duplicate.

The more defensive check is the one the codebase already settled on — identify by the L1's `cfnResourceType`, per ADR-0011 §3 and `isCfnAlarm` in `@composurecdk/cloudwatch`:

```ts
function findExistingLogGroup(stack: Stack): LogGroup | undefined {
  const existing = stack.node.tryFindChild(DELEGATION_PROVIDER_LOG_GROUP_ID);
  const l1 = existing?.node.defaultChild;
  if (!CfnResource.isCfnResource(l1)) return undefined;
  if (l1.cfnResourceType !== CfnLogGroup.CFN_RESOURCE_TYPE_NAME) return undefined;
  return existing as LogGroup;
}
```

`CfnResource.isCfnResource` and `CFN_RESOURCE_TYPE_NAME` both predate the `^2.225.0` floor, so this is stable across the whole peer range. There's a test that stands a non-`LogGroup` construct with a `CfnLogGroup` default child at the dedup id — it fails against the `instanceof` form and passes against this one.

One thing worth your call, outside this diff: `query-logging.ts:119` has the same shape (`existing instanceof ResourcePolicy`) in `ensureSharedResourcePolicy`, with the same latent behaviour. I left it alone to keep this PR scoped. Happy to fix it here or file it separately — say which you prefer.

---

packages/route53/src/defaults.ts
laazyj: Agreed — they fail the test for what `defaults.ts` is for. Neither is overridable configuration; they're values the builder derives the log group from, and the log group itself is what reaches the consumer, via `result.providerLogGroup`. Fixed in 2c267ce.

Both now live in `cross-account-delegation-provider-logging.ts` next to their only use site, and both are off the package barrel. `CROSS_ACCOUNT_ZONE_DELEGATION_DEFAULTS` stays in `defaults.ts` and stays exported — `ttl`, `removalPolicy`, and `providerLogging` are all overridable through the fluent API, so it earns its place in the defaults table.

They keep module-level `export` so the tests can import them, matching how `QUERY_LOGGING_RESOURCE_POLICY_ID` is handled — package-internal, deliberately absent from `index.ts`. Both JSDoc blocks now say that explicitly, so the next reader doesn't have to infer it from the barrel.

---

packages/route53/src/cross-account-delegation-provider-logging.ts
laazyj: Raise a new bug describing the problem in ensureSharedResourcePolicy and the recommended fix.

laazyj: Filed as #384 — the problem in `ensureSharedResourcePolicy`, the impact (falls through to a second `ResourcePolicy` at the same construct id, so a hard synth failure rather than a duplicated policy), the recommended `cfnResourceType` fix, and a test that fails against the current guard.

While checking, I found `query-logging.ts:119` is the last `instanceof` dedup of this kind in library `src/`, but `@composurecdk/iam` has two uses of `s instanceof StatementBuilder` (`role-builder.ts:192`, `managed-policy-builder.ts:67`). Those fail differently and worse — a `StatementBuilder` from the other realm falls to the `: s` branch and gets handed to CDK as though it were a `PolicyStatement`. They'd want the `Symbol.for` brand `Ref` uses rather than the L1 idiom, since they're our own types. I noted it at the bottom of #384 rather than widening its scope; say the word and I'll file it on its own.

---

laazyj: File the iam problems as a separate bug

laazyj: Filed as #385.

Digging in to write it up, the impact is sharper than I put it above — worth flagging here since it changes the priority relative to #384.

**The `instanceof` is the dispatch that decides whether `.build()` runs**, and `StatementBuilder.build()` is where the wildcard-resource guard lives (`statement-builder.ts:144-151`). A cross-realm `StatementBuilder` takes the `: s` branch, so `.build()` never runs and `WildcardResourceError` never fires — an `Effect=Allow` on `"*"` without the explicit `allowWildcardResources(true)` opt-in passes by accident. A guard that is opt-out by design becomes opt-out by module format.

And unlike #384 there's no duplicate-id backstop to make it loud: the unbuilt builder goes into `new PolicyDocument({ statements })` as if it were a `PolicyStatement`, so it either crashes at synth with a CDK `TypeError` that names nothing in this codebase, or renders something unintended.

The exposure is also higher than I first said. Both call sites take `(PolicyStatement | StatementBuilder)[]` as public API, so `StatementBuilder` is a class the consumer constructs and hands back — the realm boundary sits directly across the API surface, and an app with an ESM entrypoint and a CJS module is enough.

Fix is the `Symbol.for` brand rather than #384's `cfnResourceType` idiom, since these are our own types rather than CDK constructs — `REF_BRAND` / `isRef` is the template. I also noted in #385 that it's worth confirming whether any other exported `@composurecdk` class crossing an API boundary needs the same brand, so that list can be closed rather than rediscovered.

---

laazyj:
Pushed `136b158` — a rebase interaction I'd have missed if I hadn't re-run the gate against the new base.

The rebase onto `59ece97` brought #375, which widened `LogGroupBuilderProps.encryptionKey` to a `Resolvable` and gave `LogGroupBuilder.build()` a `context` parameter. This builder was still calling the sub-builder as `subBuilder.build(stack, id)`, so a `providerLogging({ configure })` callback that reached for a sibling KMS key with `ref()` would resolve against an empty record and throw *"component not found"* — precisely the case the widened prop exists to allow. `context` is now threaded from the delegation record's `build()` through to the sub-builder, with a test that refs a `Key` from the build context.

Full gate re-run against the rebased base: build, lint, format, test, `check:exports`, `validate`, `cdk-floors:check` all pass. route53 is at 28 tests in the new file.

**Same gap, not fixed here:** `resolveQueryLogging` has it too (`query-logging.ts:102`), and it can't be fixed the same way — `HostedZoneBuilder.build(scope, id)` takes no `context` at all, so there is nothing to thread. Widening that signature is a change to an existing public builder rather than anything this PR introduced, so I've left it. The `lifecycle-build-context-required` lint rule doesn't catch it because `HostedZoneBuilder` itself declares no `Resolvable` prop — the resolvable is inside the sub-builder it delegates to, which the rule can't see. Worth a third issue if you want it; say the word.

---

Issue #382, opened 2026-08-15
refactor(core): ADR-0009's recurrence trigger has fired — weigh a typed default-yields-to-sibling primitive


laazyj:
Raised from the review of #375, which flagged this in its description rather than deciding it inline.

[ADR-0009](../blob/main/docs/adr/0009-defaults-yield-to-mutually-exclusive-siblings.md) closes with an explicit escalation clause:

> The fix is per-instance. If default-vs-sibling collisions recur, that is the signal to weigh a generic, typed primitive in `@composurecdk/core` rather than repeating the pattern.

#375 added three more instances at once, bringing the total to six across five packages — and, more tellingly, they are written in **three different shapes**.

## The instances

| Package | Default | Yields to | Shape |
| --- | --- | --- | --- |
| `dynamodb` | `billingMode: PAY_PER_REQUEST` | `readCapacity` / `writeCapacity` | inline `delete merged.billingMode` |
| `dynamodb` | `encryption: AWS_MANAGED` | `encryptionKey` | inline `merged.encryption = CUSTOMER_MANAGED` |
| `ec2` | VPC `maxAzs` | `availabilityZones` | (pre-existing) |
| `s3` | `encryption: S3_MANAGED` | `encryptionKey` | extracted `encryptionKeyProps()` helper |
| `sqs` | `encryption: SQS_MANAGED` | `encryptionMasterKey` | extracted `encryptionMasterKeyProps()` helper |
| `kms` | `enableKeyRotation: true` | a non-rotatable `keySpec` | inline `delete merged.enableKeyRotation` |

Two of those helpers (`s3` and `sqs`) are the same function under two names — identical signature and body, differing only in the prop name and the enum constant.

## Why it is worth deciding now rather than at instance seven

- **No canonical form to copy.** A seventh author picks one of three shapes by coin flip. The three are not equivalent either: "delete the default and let CDK infer" and "assign the customer-managed value explicitly" produce the same template today but rest on different assumptions about CDK's inference staying put.
- **The rule is subtle enough to get wrong.** Each site has to distinguish *the user set this prop* from *the default set it* — reading the user's props, not the merged object. That is the kind of thing a primitive should encode once.
- **A bug in the yield logic is currently a six-place fix.**

## What to weigh

Sketch, not a proposal — the shape is the actual question:

```ts
// in @composurecdk/core — CDK-agnostic, so no aws-cdk-lib import
yieldsTo(defaults, userProps, [
[...]

Issue #380, opened 2026-08-15
feat(neptune): widen ClusterBuilderProps.kmsKey to Resolvable so a composed CMK can be referenced


laazyj:
Follow-up to #373 / #375, which added `@composurecdk/kms` and widened the key-consuming props on the library's stateful resources to `Resolvable`.

`@composurecdk/neptune` was left out of that sweep. `ClusterBuilderProps` omits only `vpc` and `securityGroups`, so `kmsKey` still takes a concrete key and a `createKeyBuilder()` component cannot be referenced into it:

```ts
// Wanted:
createClusterBuilder().kmsKey(ref<KeyBuilderResult>("clusterKey").get("key"));
```

A Neptune cluster is exactly the kind of stateful resource where a CMK is the point — independent rotation, a key policy you control, and CloudTrail visibility of decrypt calls.

**Proposed change**

The same three-step widening the other five packages now use — `Omit` + a `Resolvable` re-declaration, resolved in `build()`:

```ts
export interface ClusterBuilderProps
  extends Omit<DatabaseClusterProps, "vpc" | "securityGroups" | "kmsKey"> {
  kmsKey?: Resolvable<IKey>;
}
```

Check whether `storageEncrypted` needs the ADR-0009 treatment here: if the builder defaults it, supplying a key should imply it rather than requiring both, the way `s3`'s `encryptionKey` now infers `BucketEncryption.KMS`.

**The wrinkle worth deciding**

Same as #379: on current `@aws-cdk/aws-neptune-alpha`, `kmsKey` is typed `kms.IKeyRef`, not `kms.IKey`. At this package's floor (2.190.0) it is `IKey`, and `IKeyRef` does not exist. `Resolvable<IKey>` compiles at the floor but narrows what a consumer on a newer CDK can pass; `Resolvable<IKeyRef>` does the reverse. Neptune also has the extra constraint that its alpha module is version-locked to its matching aws-cdk-lib release (see the `peerFloors` entry in `cdk-floors.json`), so raising the floor moves both.

**Acceptance**

- [ ] `kmsKey` accepts a concrete key and a `Ref`, resolved at build
- [ ] Test covering resolve-from-context, following the pattern in `packages/s3/test/bucket-builder.test.ts`
- [ ] `cdk-floors enforce 2.190.0` still green (or the floor and its locked alpha peer raised deliberately)
- [ ] The caveat paragraph in `packages/kms/README.md` updated to drop `neptune`

Issue #379, opened 2026-08-15
feat(lambda): widen FunctionBuilderProps.environmentEncryption to Resolvable so a composed CMK can be referenced


laazyj:
Follow-up to #373 / #375, which added `@composurecdk/kms` and widened the key-consuming props on the library's stateful resources to `Resolvable`.

`@composurecdk/lambda` was left out of that sweep. `FunctionBuilderProps` omits only `role`, so `environmentEncryption` still takes a concrete key and a `createKeyBuilder()` component cannot be referenced into it:

```ts
// Works today only if the key is a construct built outside compose.
createFunctionBuilder().environmentEncryption(someKey);

// Wanted:
createFunctionBuilder().environmentEncryption(ref<KeyBuilderResult>("envKey").get("key"));
```

**Proposed change**

The same three-step widening the other five packages now use — `Omit` + a `Resolvable` re-declaration, resolved in `build()`:

```ts
export interface FunctionBuilderProps extends Omit<FunctionProps, "role" | "environmentEncryption"> {
  environmentEncryption?: Resolvable<IKey>;
}
```

`build()` already takes `context` (it resolves `role`), so the change is confined to the props interface and one spread in the merge.

**The wrinkle worth deciding**

On current aws-cdk-lib, `FunctionProps.environmentEncryption` is typed `kms.IKeyRef`, not `kms.IKey`. At this package's floor (2.168.0) it is `IKey`, and `IKeyRef` does not exist. So:

- `Resolvable<IKey>` compiles at the floor but **narrows** what a consumer on a newer CDK can pass — anything that is an `IKeyRef` but not an `IKey` stops being accepted.
- `Resolvable<IKeyRef>` matches current CDK but does not compile at 2.168.0.

Options are to accept the narrowing (it is a pre-1.0 library and `IKey` is what essentially every caller holds), or to raise the lambda floor to whichever release introduced `IKeyRef` and widen to that. Worth deciding before implementing; #375 deliberately left this alone rather than pick silently.

**Acceptance**

- [ ] `environmentEncryption` accepts a concrete key and a `Ref`, resolved at build
- [ ] Test covering resolve-from-context, following the pattern in `packages/s3/test/bucket-builder.test.ts`
- [ ] `cdk-floors enforce` still green for the lambda floor (or the floor raised deliberately, with `cdk-floors.json` updated)
- [ ] The caveat paragraph in `packages/kms/README.md` updated to drop `lambda`

PR #383, opened 2026-08-15
feat(examples): encrypt the crud-api table with a composed customer-managed key

laazyj:
## What & why

Follow-up to #375, as requested in review: migrate an existing example to a customer-managed KMS key so the new package is proved against live AWS, and teach `cleanDeskPolicy` to tear keys down cheaply.

The gadgets table in `ComposureCDK-CrudApiStack` is now encrypted with a CMK that is itself a component — `tableKey` is a `createKeyBuilder()` sibling the table depends on by name, so the key is placed by the stack strategy and ordered by the graph rather than built imperatively before `compose`.

## What the example proves by omission

This is why it is worth deploying rather than only unit-testing:

- **No encryption mode is set.** `.encryptionKey(ref(…))` alone yields `TableEncryption.CUSTOMER_MANAGED` — the builder's `AWS_MANAGED` default is mutually exclusive with a customer key, so it yields ([ADR-0009](../blob/main/docs/adr/0009-defaults-yield-to-mutually-exclusive-siblings.md)).
- **No KMS grant is declared.** `tableGrants.readWrite` already reaches the key, because CDK's own `grantReadWriteData` extends to the table's `encryptionKey`. The updated snapshot shows `kms:Decrypt` / `kms:GenerateDataKey*` on the API Gateway role's policy with no `keyGrants` call anywhere in the stack; a test asserts exactly that. Adding a `keyGrants` call here would have been redundant permission, not extra safety.

## One structural change, and why

The table moves from `TableV2` to the classic `Table`. `TableV2` encodes SSE per replica, which CDK refuses to render in a region-agnostic stack:

```
Replica SSE specification cannot be rendered in a region agnostic stack
```

Every example here is region-agnostic so CI can deploy it to whatever region it runs in. Keeping `TableV2` would mean binding this one stack to an explicit `env` — making it the only example that cannot deploy anywhere, and putting a concrete account/region in its snapshot. I verified the three combinations by synthesising each:

| Combination | Region-agnostic stack |
| --- | --- |
| `TableV2` + CMK | **fails** |
| classic `Table` + CMK | OK |
| SQS + CMK | OK |

The example carries a comment explaining this, so the next person doesn't rediscover it. Note this is a **resource replacement** on the deployed sandbox stack (`AWS::DynamoDB::GlobalTable` → `AWS::DynamoDB::Table`), which is fine for an ephemeral example but worth knowing before the next `deploy-test` run.

## cleanDeskPolicy

Two new injectors:

- **KMS keys** — `DESTROY` plus a 7-day pending-deletion window. Seven is the floor, not a choice: AWS enforces a 7-to-30 day waiting period on `ScheduleKeyDeletion` and offers no immediate-delete option, so "immediately expire deleted keys" is not reachable and a torn-down sandbox key still bills for a week. The examples README now says so under Costs, since it is the one resource there outside the free tier.
- **Classic `Table`** — the existing injector keys on `TableV2`'s construct id, so the classic table needed its own to get `DESTROY` + `deletionProtection: false`.

## CI permissions

No change to `.github/cloudformation/github-oidc-role.yml`, and none needed: the smoke test drives the API and never calls KMS itself, and resource creation runs under the CDK bootstrap execution role rather than the deploy-test OIDC role. No hand-deployed stack needs redeploying before the next `deploy-test` run.

## Checklist

- [x] Linked to an issue — follow-up to #373 / #375 requested in review
- [x] `npm run verify` passes locally, except `actionlint`, whose shellcheck download is blocked by this environment's egress proxy (403). No workflow files are touched. Every other gate passes: `format:check`, `build`, `catalogue:check`, `check:exports`, `lint`, `cdk-floors:check`, `validate`, `test`.
- [x] Tests added/updated — key/alias/SSE assertions, the grant-reaches-the-key assertion, the clean-desk key teardown test, and an updated snapshot
- [x] No new example stack added (this migrates an existing one)

Also raised from the #375 review: #379 (lambda `environmentEncryption`), #380 (neptune `kmsKey`), and #382 (ADR-0009's recurrence trigger).

---
_Generated by [Claude Code](https://claude.ai/code/session_016ZQt2jXkmqnaNr9mT93APU)_


laazyj:
Deployed to sandbox and confirmed working smoke tests.

PR #387, opened 2026-08-15
feat(lambda): widen environmentEncryption to Resolvable so a composed CMK can be referenced

### LINKED ISSUE #379: feat(lambda): widen FunctionBuilderProps.environmentEncryption to Resolvable so a composed CMK can be referenced

laazyj:
## What & why

`@composurecdk/lambda` was the one builder left out of #375's sweep, so a function encrypting its environment variables with a customer-managed key still had to take that key as a construct built before `compose` — outside the dependency map, stack placement and build ordering the rest of the system gets. It now takes the same `Omit` + `Resolvable` re-declaration + resolve-in-`build()` shape as `s3`, `sns`, `sqs`, `dynamodb` and `logs`:

```ts
createFunctionBuilder().environmentEncryption(ref<KeyBuilderResult>("envKey").get("key"));
```

### The `IKey`/`IKeyRef` wrinkle #375 deferred

AWS is migrating key-consuming props from `kms.IKey` to the wider `kms.IKeyRef` ([reference interfaces](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_kms.IKeyRef.html), aws-cdk-lib 2.215.0). `FunctionProps.environmentEncryption` moved in exactly that release — verified against the CDK source at each tag, 2.214.0 has `IKey` and 2.215.0 has `IKeyRef` — while this package's floor is 2.168.0, where `IKeyRef` does not exist. Neither interface can be named without the builder disagreeing with the CDK it is installed against: `IKey` rejects an `IKeyRef` that CDK itself accepts, and `IKeyRef` does not compile at the floor.

So the re-declared prop names neither, and reads the type from the consumer's own aws-cdk-lib:

```ts
environmentEncryption?: Resolvable<NonNullable<FunctionProps["environmentEncryption"]>>;
```

This is what the surrounding `Omit<FunctionProps, …>` already does for every prop the builder does not re-declare — the widening now tracks upstream for this one too, resolving to `Resolvable<IKey>` at the floor and `Resolvable<IKeyRef>` above 2.215.0. Verified by typechecking the declaration against real installs at both ends (2.168.0 and 2.263.0), accepting a concrete `Key`, an `IKey`, and a `Ref` to either.

The alternative — raising the floor to 2.215.0 — would cost lambda consumers ten months of CDK releases (2.168.0 shipped 2024-11-20) to accept a type no caller in this library produces. **The lambda floor is unchanged at 2.168.0**, `cdk-floors check` is green, and no new aws-cdk-lib API is named.

### `logs` had already picked up the narrowing

`LogGroupProps.encryptionKey` moved to `IKeyRef` in 2.215.0 as well, so #375's `Resolvable<IKey>` had already narrowed it for consumers above that release. It gets the same treatment here. The other four widened props (`s3`, `sns`, `sqs`, `dynamodb`) are still `IKey` upstream and keep the literal spelling until CDK moves them.

### Guarding it

Nothing in CI defends a floor-gated *type*: `cdk-floors enforce` runs the unit suites, which vitest does not typecheck, and `build`/`typecheck` run against the latest devDependency — so "simplifying" this back to `IKey` would break only in a consumer's own `tsc`. Two cheap guards, both where a maintainer already looks:

- `cdk-floors.json` records the constraint on the `lambda` and `logs` entries, in the shape the `kms` entry already uses for `IKey.grantSign`.
- `docs/architecture.md`'s Ref-support recipe now says to take `T` from the CDK prop when the widening is an `Omit`-and-re-declare — that recipe is what the next key-prop widening will follow, and CDK is mid-migration.

No ADR: per [when to write an ADR](docs/adr/README.md#when-to-write-an-adr) this is how one prop type is spelled in two packages, not a change to the library's shape.

## Checklist

- [x] Linked to an issue (or it's a small, obvious fix) — Closes #379
- [ ] `npm run verify` passes locally — every stage passes except `actionlint`, which cannot run in this environment: the pinned `shellcheck` devDependency downloads its binary at install time and the network here returns 403. No workflow files are touched by this PR. `format:check`, `build`, `catalogue:check`, `check:exports`, `lint`, `cdk-floors:check`, `validate` and `test` were each run and are green.
- [x] Tests added/updated for the change — concrete-key and resolve-from-context cases for `environmentEncryption`, following `packages/s3/test/bucket-builder.test.ts`
- [x] If it adds an example stack — n/a

Closes #379

---
_Generated by [Claude Code](https://claude.ai/code/session_015kWvJ3e7UfPg8Uwmge4WpZ)_

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -414,8 +414,6 @@ class MyBuilder implements Lifecycle<MyResult> {
 
 This is the only change required. The builder does not need to know whether it received a concrete value or a `Ref` — `resolve` handles both uniformly.
 
-When the widened prop is an `Omit`-and-re-declare of a CDK prop, take `T` from the CDK prop itself — `Resolvable<NonNullable<FunctionProps["environmentEncryption"]>>` — rather than naming the interface CDK happens to use today. Re-declaring is the one place a builder stops tracking the consumer's `aws-cdk-lib`, and CDK's prop types move: the key-consuming props are migrating from `kms.IKey` to the wider `kms.IKeyRef` one release at a time, and a hardcoded `Resolvable<IKey>` starts rejecting keys the underlying CDK accepts the moment that prop moves. Reading the type keeps the builder honest above and below the change, and compiles at floors that predate the new interface. See the table in [`@composurecdk/kms`](../packages/kms/README.md#keys-as-components) for which props this currently affects.
-
 ## Policies
 
 Some concerns are not builder-local: they apply to whatever matching constructs happen to exist under a scope, not to any one component. Examples: attaching CloudWatch alarm actions to every `Alarm` in a stack, overriding `removalPolicy` on every stateful resource for a throwaway environment, or enforcing a tagging convention across services.
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -414,6 +414,8 @@ class MyBuilder implements Lifecycle<MyResult> {
 
 This is the only change required. The builder does not need to know whether it received a concrete value or a `Ref` — `resolve` handles both uniformly.
 
+When the widened prop is an `Omit`-and-re-declare of a CDK prop, take `T` from the CDK prop itself — `Resolvable<NonNullable<FunctionProps["environmentEncryption"]>>` — rather than naming the interface CDK happens to use today. Re-declaring is the one place a builder stops tracking the consumer's `aws-cdk-lib`, and CDK's prop types move: the key-consuming props are migrating from `kms.IKey` to the wider `kms.IKeyRef` one release at a time, and a hardcoded `Resolvable<IKey>` starts rejecting keys the underlying CDK accepts the moment that prop moves. Reading the type keeps the builder honest above and below the change, and compiles at floors that predate the new interface. See the table in [`@composurecdk/kms`](../packages/kms/README.md#keys-as-components) for which props this currently affects.
+
 ## Policies
 
 Some concerns are not builder-local: they apply to whatever matching constructs happen to exist under a scope, not to any one component. Examples: attaching CloudWatch alarm actions to every `Alarm` in a stack, overriding `removalPolicy` on every stateful resource for a throwaway environment, or enforcing a tagging convention across services.
--- a/packages/kms/README.md
+++ b/packages/kms/README.md
@@ -30,12 +30,12 @@ The key-consuming props on the library's stateful resources accept a `Resolvable
 | `@composurecdk/dynamodb` | `encryptionKey`         | `Resolvable<IKey>` (classic `Table`) |
 | `@composurecdk/sqs`      | `encryptionMasterKey`   | `Resolvable<IKey>`                   |
 | `@composurecdk/sns`      | `masterKey`             | `Resolvable<IKey>`                   |
-| `@composurecdk/logs`     | `encryptionKey`         | `Resolvable<IKey>` \*                |
-| `@composurecdk/lambda`   | `environmentEncryption` | `Resolvable<IKey>` \*                |
+| `@composurecdk/logs`     | `encryptionKey`         | `Resolvable<IKey>`                   |
+| `@composurecdk/lambda`   | `environmentEncryption` | `Resolvable<IKey>`                   |
 
-\* These two accept more than an `IKey`, because CDK is migrating key-consuming props to the wider [`kms.IKeyRef`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_kms.IKeyRef.html) (aws-cdk-lib 2.215.0) and these are the two that have moved. Naming either interface would put the builder at odds with the CDK it is installed against — `IKey` would reject an `IKeyRef` that CDK itself accepts, and `IKeyRef` does not exist at those packages' floors — so their inner type is read from the consumer's own aws-cdk-lib (`Resolvable<NonNullable<FunctionProps["environmentEncryption"]>>`) and tracks the migration in both directions. Passing an `IKey` or a `Ref` to a key builder is unaffected. The `keyGrants` helpers below still take `Resolvable<IKey>`, so a bare `IKeyRef` reaches those two props but not a grant.
+An `IKey` is what these props are written against, but it is not the limit of what they take: where CDK has widened a prop to the broader [`kms.IKeyRef`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_kms.IKeyRef.html), the builder's prop is read from CDK's own type rather than pinned to an interface, so it accepts whatever the `aws-cdk-lib` you have installed accepts. The `keyGrants` helpers below are pinned to `IKey`, so a key that is only an `IKeyRef` reaches a resource prop but not a grant.
 
-One key-consuming prop elsewhere still takes a concrete key — `@composurecdk/neptune`'s `kmsKey` ([#380](https://github.com/laazyj/composureCDK/issues/380)). Build the key as a component and pass `result.key` to it until it is widened.
+For a key-consuming prop not listed above, build the key as a component and pass `result.key` to it.
 
 ```ts
 import { AttributeType, TableEncryptionV2 } from "aws-cdk-lib/aws-dynamodb";
--- a/packages/kms/README.md
+++ b/packages/kms/README.md
@@ -30,10 +30,10 @@ The key-consuming props on the library's stateful resources accept a `Resolvable
 | `@composurecdk/dynamodb` | `encryptionKey`         | `Resolvable<IKey>` (classic `Table`) |
 | `@composurecdk/sqs`      | `encryptionMasterKey`   | `Resolvable<IKey>`                   |
 | `@composurecdk/sns`      | `masterKey`             | `Resolvable<IKey>`                   |
-| `@composurecdk/logs`     | `encryptionKey`         | `Resolvable` of the CDK prop type    |
-| `@composurecdk/lambda`   | `environmentEncryption` | `Resolvable` of the CDK prop type    |
+| `@composurecdk/logs`     | `encryptionKey`         | `Resolvable<IKey>` \*                |
+| `@composurecdk/lambda`   | `environmentEncryption` | `Resolvable<IKey>` \*                |
 
-Two of those rows do not name `IKey`. CDK is migrating key-consuming props from `kms.IKey` to the wider `kms.IKeyRef` ([reference interfaces](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_kms.IKeyRef.html), aws-cdk-lib 2.215.0); `LogGroupProps.encryptionKey` and `FunctionProps.environmentEncryption` have already moved, while the other four have not. Those two props take `Resolvable<NonNullable<LogGroupProps["encryptionKey"]>>` and `Resolvable<NonNullable<FunctionProps["environmentEncryption"]>>` — the inner type is read from the consumer's own aws-cdk-lib, so the builder accepts exactly what CDK accepts at whatever version is installed, above and below the migration. A concrete `IKey` and a `Ref` to a key builder satisfy both spellings, so the wiring below is the same either way.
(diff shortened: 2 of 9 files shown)


packages/kms/README.md
```diff
@@ -23,16 +23,19 @@ The point of the builder is that a CMK stops being a construct you create impera
...
+| `@composurecdk/lambda`   | `environmentEncryption` | `Resolvable` of the CDK prop type    |
+
+Two of those rows do not name `IKey`. CDK is migrating key-consuming props from `kms.IKey` to the wider `kms.IKeyRef` ([reference interfaces](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_kms.IKeyRef.html), aws-cdk-lib 2.215.0); `LogGroupProps.encryptionKey` and `FunctionProps.environmentEncryption` have already moved, while the other four have not. Those two props take `Resolvable<NonNullable<LogGroupProps["encryptionKey"]>>` and `Resolvable<NonNullable<FunctionProps["environmentEncryption"]>>` — the inner type is read from the consumer's own aws-cdk-lib, so the builder accepts exactly what CDK accepts at whatever version is installed, above and below the migration. A concrete `IKey` and a `Ref` to a key builder satisfy both spellings, so the wiring below is the same either way.
+
+One key-consuming prop elsewhere still takes a concrete key — `@composurecdk/neptune`'s `kmsKey` ([#380](https://github.com/laazyj/composureCDK/issues/380)). Build the key as a component and pass `result.key` to it until it is widened.
```
laazyj: I don't like this living in the README - it requires a backward update to the doc when the update lands. I want the README to be stable for both the current state and the target state (sacrificing some nuance in current state if necessary).

This also raises a bigger question on whether this _actually is an ADR_. The "Omit and redeclare" pattern is already an established pattern for props widening and this `IKey` --> `IKeyRef` transition has identified a real future-proofing/backward-compatibility problem with the established approach. Consider the following:
1. What would an ADR that established a more robust approach to "Omit + redeclare" for builder props would communicate?
2. What would an implementation pass to apply the new pattern consistently across ALL packages look like - are there any risks to it?

packages/lambda/README.md
```diff
@@ -49,6 +49,31 @@ const handler = createFunctionBuilder()
...
+      .code(Code.fromAsset("lambda"))
+      .environmentEncryption(ref<KeyBuilderResult>("envKey").get("key")),
+  },
+  { envKey: [], checkout: ["envKey"] },
+);
+```
+
+CDK grants the function's execution role decrypt on a key it can see, so no extra grant is needed for the function to read its own variables.
+
+The prop's inner type is read from the consumer's own aws-cdk-lib (`NonNullable<FunctionProps["environmentEncryption"]>`) rather than named as `IKey`, because CDK widened `FunctionProps.environmentEncryption` from `kms.IKey` to `kms.IKeyRef` in aws-cdk-lib 2.215.0. Naming either interface would make the builder disagree with the CDK it is installed against — `IKey` would reject an `IKeyRef` that CDK itself accepts, and `IKeyRef` does not exist at this package's 2.168.0 floor.
```
laazyj: This clause (L75) is implied and doesn't need to be documented here.

packages/logs/README.md
```diff
@@ -62,3 +62,5 @@ compose(
   { logKey: [], audit: ["logKey"] },
 );
 ```
+
+The prop's inner type is read from the consumer's own aws-cdk-lib (`NonNullable<LogGroupProps["encryptionKey"]>`) rather than named as `IKey`, because CDK widened `LogGroupProps.encryptionKey` from `kms.IKey` to `kms.IKeyRef` in aws-cdk-lib 2.215.0. Naming either interface would make the builder disagree with the CDK it is installed against — `IKey` would reject an `IKeyRef` that CDK itself accepts, and `IKeyRef` does not exist at this package's 2.1.0 floor.
```
laazyj: This compatibility is implied and doesn't need to be restated in the README

laazyj (commented):
Consider whether this actually warrants an architectural change

packages/kms/README.md
laazyj: README restructured in 73caa28 — the table no longer names which props have moved, one sentence covers "not limited to `IKey` where CDK has widened it", and an unlisted prop is handled by a general instruction instead of a per-package exception. Nothing there needs a backward edit as the migration proceeds. The two implied clauses are dropped.

On the bigger question — you're right, and I had the framing wrong. I scoped it as "how one prop type is spelled in two packages" and concluded no ADR. Scoped as **"the established Omit-and-redeclare pattern silently narrows the builder's public API whenever CDK widens a prop"**, it is a different decision and it clears the bar. Evidence for that reframing, gathered against the installed aws-cdk-lib 2.263.0:

| Builder prop | We declare | CDK declares today |
| --- | --- | --- |
| `events` `eventBus` | `Resolvable<IEventBus>` | `IEventBusRef` |
| `cloudfront` `certificate` | `Resolvable<ICertificate>` | `ICertificateRef` |
| `logs` `encryptionKey` | (fixed here) | `IKeyRef` |
| `lambda` `environmentEncryption` | (this PR) | `IKeyRef` |
| `neptune` `kmsKey` | (#389) | `IKeyRef` |

`IEventBus extends IResource, IEventBusRef` and `ICertificate extends IResource, ICertificateRef`, so **`events` and `cloudfront` are narrowing their public API right now**, and nobody noticed. This was never a KMS issue — KMS is just where it happened to surface. That, more than anything, is the argument for an ADR: the pattern has a systematic defect and the current documentation tells the next author to reproduce it.

## 1. What the ADR would communicate

**The failure mode, stated once.** Re-declaring is the single place a builder stops tracking the consumer's `aws-cdk-lib`. Everything inside `Omit<XProps, K>` follows the installed CDK; the re-declared key freezes at whatever interface was current when it was written. CDK's prop types are a moving target — the reference-interface migration (2.215.0) is the current instance, not the last one — so the frozen spelling drifts from the CDK it is installed against, and always in the direction of rejecting values CDK accepts.

**Why no gate catches it.** `cdk-floors enforce` runs the unit suites, and vitest does not typecheck. `build`/`typecheck` run against the latest devDependency, where the narrow spelling still compiles — it is narrower, not wrong. The failure surfaces only in a consumer's own `tsc`, as "your builder rejects the key my CDK told me to pass." An entire failure class with no defence, which is exactly the kind of thing an ADR exists to record.

**The rule.** When a prop is re-declared *only* to widen it to `Resolvable`, take the inner type from the CDK prop: `Resolvable<NonNullable<XProps["k"]>>`. When it is re-declared to change its *shape*, the builder's own type is the point and the rule does not apply.

That distinction is load-bearing, and it splits the surface cleanly. Of 33 re-declared props across 15 packages, the shape-changers are a real minority with a clear signature — `s3` `serverAccessLogsBucket`/`Prefix` → `ServerAccessLogsConfig`, `ec2` `flowLogs` → `FlowLogsConfig`, `route53` `queryLogsLogGroupArn` → `QueryLoggingConfig`, `cloudfront` `functionAssociations` → `functions`, the dynamodb event source's optionality change. The rest are pure widenings.

**Two exclusions worth naming**, because a blanket rule gets them wrong:

- **`any`-typed props.** `custom-resources` declares `parameters?: Resolvable<Record<string, unknown>>` where CDK has `parameters?: any`. Tracking would yield `Resolvable<any>` — the rule would actively destroy type safety.
- **Primitives.** `route53` `values` (`string[]`) will never migrate to a reference interface. Tracking is noise. Scope the rule to resource-reference types.

**Enforcement, honestly.** Since types are invisible to the test gate, the ADR should say what does defend it. A lint rule in `@composurecdk/eslint-plugin` is tractable *because it is syntactic, not semantic*: flag `Resolvable<SomeNamedInterface>` on a property whose name appears in the `Omit<..., K>` of the same interface's `extends`. No type resolution needed, so it works in the existing flat-config setup, and it fires at authoring time — which is the only moment anyone is thinking about this.

**The accepted cost.** The prop's type reads as an indexed access rather than a name, so hover text and generated docs get less literal. Package READMEs keep describing it as `Resolvable<IKey>`, which is what a user passes. And it must stay *inline* — a named alias (`type EnvKey = …`) would put an unnameable type in the emitted mapped type and reintroduce ADR-0001's TS2883 for consumers writing `createSystem()`.

## 2. What the sweep looks like, and its risks

Roughly: triage the 33 re-declarations into widen-only vs shape-changing, apply the rule to the widen-only set, add the lint rule, and delete the manifest notes that only exist because nothing else defends this.

Risks, in the order I'd worry about them:

1. **It is a public API change in ~15 packages at once.** `*BuilderProps` is public (ADR-0001), so anyone who wrote `FunctionBuilderProps["environmentEncryption"]` sees a different type. Pre-1.0, and it only ever widens what is accepted, but it is semver-relevant and lands everywhere simultaneously.
2. **Per-package floor verification.** "Compiles at the floor" is a per-package claim: the indexed access is only safe where the prop *exists* at that package's floor. I verified lambda (2.168.0), logs (2.1.0) and neptune (2.190.0-alpha.0) against real installs; a sweep needs the same for each package touched, and floors range from 2.1.0 to 2.225.0.
3. **A mechanical diff hides the real fixes.** 30-odd near-identical hunks, of which two (`events`, `cloudfront`) are actual bug fixes that deserve their own scrutiny and possibly their own tests. Worth landing those two separately, before or after, rather than burying them.
4. **`any` and primitive props** — per the exclusions above; getting these wrong is worse than leaving them alone.
5. **Silent regression afterwards.** Without the lint rule the sweep decays: nothing stops the next author writing `Resolvable<IKey>`, and it will still compile. The rule is what makes the sweep durable, so I would not do the sweep without it.

The `attw`/`publint` gate and `check:exports` give some cover on the emission side, but none of it sees narrowing, so it does not change the picture.

I can write the ADR next, with `lambda`/`logs` as its first application here and the sweep as a follow-up — or fold the sweep in, if you would rather see the whole thing at once. My preference is the split, mainly for risk 3.

---

laazyj: Good. We will tackle this bigger job separately and ship the narrower fix in the meantime.

docs/architecture.md
```diff
@@ -414,6 +414,8 @@ class MyBuilder implements Lifecycle<MyResult> {
 
 This is the only change required. The builder does not need to know whether it received a concrete value or a `Ref` — `resolve` handles both uniformly.
 
+When the widened prop is an `Omit`-and-re-declare of a CDK prop, take `T` from the CDK prop itself — `Resolvable<NonNullable<FunctionProps["environmentEncryption"]>>` — rather than naming the interface CDK happens to use today. Re-declaring is the one place a builder stops tracking the consumer's `aws-cdk-lib`, and CDK's prop types move: the key-consuming props are migrating from `kms.IKey` to the wider `kms.IKeyRef` one release at a time, and a hardcoded `Resolvable<IKey>` starts rejecting keys the underlying CDK accepts the moment that prop moves. Reading the type keeps the builder honest above and below the change, and compiles at floors that predate the new interface. See the table in [`@composurecdk/kms`](../packages/kms/README.md#keys-as-components) for which props this currently affects.
```
laazyj: Drop this. It will land in a follow-up as an ADR with fuller description and justification.

laazyj (commented):
I don't want the change to architecture.md in this PR

docs/architecture.md
laazyj: Dropped in 45d366c. The recipe is back to what it was, and the general rule lands with the ADR.

I kept the `gatedBy` notes in `cdk-floors.json` for `lambda` and `logs` — they are specific to the two floors this PR changes behaviour at, rather than general guidance, and the manifest is where someone lowering a floor will actually meet the constraint. Say the word if you would rather those went with the ADR too, and this PR carried nothing but the widening.

---

